PR #5, opened 2019-02-27
[MegaMenu] Fix outline on focus of "View All in [hub name]"

ncksllvn:
This PR fixes an issue with the focus outline of the "View All in [hub name]" stretching beyond the width of the MegaMenu.

Ticket - https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14306

### Before fix
image

### After fix
image

### DIFF
--- a/packages/formation/sass/modules/_m-megamenu.scss
+++ b/packages/formation/sass/modules/_m-megamenu.scss
@@ -18,7 +18,6 @@ $marketing-container-height: 380px;
     a {
       padding: 20px 25px 20px 7px;
       font-weight: bold;
-      width: 111%;
 
       svg {
         width: 15px;


ncksllvn:
I think the intent was for the outline to stretch across the full width of the MegaMenu, which wasn't easily achievable because `100%` wouldn't account for the parent element's padding. I think a better thing to do here is just omit the width property and let the width default to auto. I think this makes sense from a UI perspective too because the `>` seems like it should be at the end of the element.

packages/formation/sass/modules/_m-megamenu.scss
```diff
@@ -18,7 +18,6 @@ $marketing-container-height: 380px;
     a {
       padding: 20px 25px 20px 7px;
       font-weight: bold;
-      width: 111%;
```
ncksllvn: @accbjt Do you know of any side effects that removing this will have?

jbalboni: Does this look ok on mobile? That'd be my only guess.

ncksllvn: It doesn't appear to have any impact on mobile, because the width of the list item already extends past the width of the container lol.

image

PR #8, opened 2019-03-01
Tweaks to alerts and alert modals

U-DON:
* Refactored alert modals to use alert classes.
* Support for 'continue' status for alert modals.
* Alerts and modals share common close button.
* Spacing changes for alerts (and alert modals) to more closely match design.
* Corrected `formation` peer dep version for `formation-react`.

### DIFF
--- a/packages/formation-react/src/components/modal/Modal.jsx
+++ b/packages/formation-react/src/components/modal/Modal.jsx
@@ -2,6 +2,8 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import classNames from 'classnames';
 
+import AlertBox from '../alerts/AlertBox/AlertBox';
+
 const ESCAPE_KEY = 27;
 
 class Modal extends React.Component {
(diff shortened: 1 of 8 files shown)


jaredcunha-usds:
Hi - I have a couple of small suggestions: 

**1) Instead of hardcoding margins and padding values, can we use the spacing units?**
We are trying to keep consistent spacing rhythms using the [8pt system](https://department-of-veterans-affairs.github.io/vets-design-system-documentation/design/spacing-units.html#why-multiples-of-8)
 
Here's some info about the spacing units: 
https://department-of-veterans-affairs.github.io/vets-design-system-documentation/design/spacing-units.html

There are functions that outputs to `rems` (default) or `px`:
https://department-of-veterans-affairs.github.io/vets-design-system-documentation/design/spacing-units.html#using-the-spacing-tokens

**2) It looks like the close button is covering some of the text in the heading. **
I'm concerned that this could be more pronounced on mobile devices. Perhaps the [padding-right utility](https://department-of-veterans-affairs.github.io/vets-design-system-documentation/utilities/padding.html#padding-right) might be useful here.

**3) Close button mixin**
Super nitpicky on my end, but the name of the mixin might. be a little broad if I saw it, and has some properties in it that depend on parent containers having positioning applied.  Suggest changing from `close-button` to `modal-close-button`?

U-DON:
@jaredcunha-usds:

1. Done! It doesn't seem anything was using the `units` function yet, so I had to import to the `shared-variables` so that it can be used across other stylesheet modules. Not sure if this is too general of an import.

2. You may be looking at the "before" example. In that one, the close button is indeed covering the header text. The "after" example shows that it can no longer overlap with any text due to the spacing changes I've made.

3. Done! Agreed that it was ambiguous. I couldn't think of a name that would apply to both modals and alerts, but settling on just one of them will be fine.

jaredcunha-usds:
@U-DON Doh! Somehow I jumbled the before/after shots. 

Regarding the naming of the close button mixin, I think it works. The pattern is merely an alert that lives inside a modal, so the button closes the modal and the alert content within it.

U-DON:
@jaredcunha-usds: The close button applies not only to alerts that live inside modals, but also alerts that are just dismissable and not in a modal context. The example displayed happens to be of an alert in a modal.

@jbalboni @accbjt Do these changes look good to y'all?

packages/formation/sass/modules/_m-modal.scss
```diff
@@ -52,28 +48,29 @@
...
+
   &-body {
     overflow-wrap: break-word;
-    padding: 2rem;
+    padding: units(2.5);
     word-break: break-word;
     word-wrap: break-word;
   }
 
-  &-button-group {
```
U-DON: Doesn't seem to be used anywhere anymore.

packages/formation/sass/modules/_m-alert.scss
```diff
@@ -205,44 +187,3 @@ p.usa-alert-heading {
...
-  .usa-alert {
-    & + .rx-app-title {
-      margin-top: 2.5rem;
-    }
-  }
-}
-
-// education benefits
-
-// this can probably be removed
```
U-DON: I checked the places where these next two rules are used, and removing the styles resulted in a very slight spacing difference. Even though they're really minor and could probably just be removed, I'll just preserve them and move them into the `edu-benefits` stylesheets when I update the dependencies in `vets-website`.

packages/formation-react/src/components/modal/Modal.jsx
```diff
@@ -99,37 +99,41 @@ class Modal extends React.Component {
...
+        <div className={wrapperClass} ref={el => { this.element = el; }}>
           {closeButton}
-          <div className="va-modal-body" role="document">
-            {title && <h3 id={titleId} className="va-modal-title">{title}</h3>}
-            <div>
-              {this.props.contents || this.props.children}
+          <div className={bodyClass}>
+            <div role="document">
+              {title && <h3 id={titleId} className={titleClass}>{title}</h3>}
+              {content && <div className={contentClass}>{content}</div>}
```
U-DON: This refactor just matches the structure to the one used for alerts. The presence of a `status` prop will just toggle each section to use alert classes rather than modal classes.

PR #11, opened 2019-03-06
New sidebar layout and stubbed information architecture

jbalboni:
This updates the sidebar to follow the information architecture we've laid out. It also moves some files around and stubs out the basic sections of the site.

screen shot 2019-03-06 at 10 05 28 am


packages/documentation/src/components/sidebar/SidebarItems.jsx
```diff
@@ -0,0 +1,76 @@
...
+              <ComponentList/>
+            </li>
+          );
+        }
+
+        if (item.href) {
+          return <li key={item.href}><Link to={item.href}>{item.name}</Link></li>;
+        }
+
+        return <li key={item.href}>{item.name} (not ready)</li>;
```
rianfowler: will `item.href` be defined here?

packages/documentation/src/components/sidebar/SidebarItems.jsx
```diff
@@ -0,0 +1,76 @@
...
+              <SidebarItems items={item.items}/>
+            </li>
+          );
+        }
+
+        if (item.query === 'componentList') {
+          return (
+            <li key={item.name}>
+              <h4>{item.name}</h4>
+              <ComponentList/>
```
rianfowler: Not sure if this would be better but we could wrap conditionals around the specific differences here: 

```
{item.href && <li key={item.href}><Link to={item.href}>{item.name}</Link></li>}
{!item.href && 
    <li key={item.name}>
      <h4>{item.name}</h4>
        {item.items && <SidebarItems/>}
        {item.query && <ComponentList/>}
     </li>
}
```

rianfowler: Could use an `item.href ? :` but I think the not `item.href` lines up more nicely.

jbalboni: I think that generally works better and updated it, but it does leave you at the end with a longer conditional for the final one, if it's a sort of catch all, display if none of the other conditions are hit, thing. I still think the approach you suggested is clearer, though.

PR #12, opened 2019-03-06
add docsearch stuff

rianfowler:
implements some of the recommendations from here: https://community.algolia.com/docsearch/recommended-configuration.html

### DIFF
--- a/packages/documentation/src/components/seo.js
+++ b/packages/documentation/src/components/seo.js
@@ -23,6 +23,14 @@ function SEO({ description, lang, meta, keywords, title }) {
                 content: metaDescription,
               },
               {
+                name: 'docsearch:language',
+                content: 'en'
+              },
+              {
+                name: 'docsearch:version',
+                content: '1.0.0'
+              },
+              {
                 property: 'og:title',
                 content: title,
               },
--- a/packages/documentation/src/layouts/layout.js
+++ b/packages/documentation/src/layouts/layout.js
@@ -136,6 +136,14 @@ class Layout extends React.Component {
                     'Template for creating design system documentatation',
               },
               {
+                name: 'docsearch:language',
+                content: 'en'
+              },
+              {
+                name: 'docsearch:version',
+                content: '1.0.0'
+              },
+              {
                 name: 'keywords',
                 content: 'design system, style guide, documentation',
               },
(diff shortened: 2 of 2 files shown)


packages/documentation/src/layouts/layout.js
```diff
@@ -154,7 +162,7 @@ class Layout extends React.Component {
             siteTitle={data.site.siteMetadata.title}
             allMarkDown={data.allMarkDown.edges}/>'
 
-            '<div className="ContentArea">{children}</div>'
+            '<div className="ContentArea docSearch-content">{children}</div>'
```
rianfowler: screen shot 2019-03-06 at 08 15 18

packages/documentation/src/layouts/layout.js
```diff
@@ -135,6 +135,14 @@ class Layout extends React.Component {
                 content:
                     'Template for creating design system documentatation',
               },
+              {
```
rianfowler: So these are just examples of meta tags we can put on pages. This part of the documentation suggests to me that we can use the meta tags to influence the way the page is indexed in the search. 

I'm putting this in the layout for now but wondering how we would organize this for page specific content. Would we have the layout query metadata (not sure what props are available on `data.site`) from the child document or push this `<Helmet>` down into the children?
screen shot 2019-03-06 at 08 15 59

packages/documentation/src/components/seo.js
```diff
@@ -22,6 +22,14 @@ function SEO({ description, lang, meta, keywords, title }) {
                 name: 'description',
                 content: metaDescription,
               },
+              {
```
rianfowler: Do we even use this seo component?

packages/documentation/src/layouts/layout.js
rianfowler: This will be used by the config to determine organize the content by levels. I'm not exactly sure how to follow this recommendation: 
screen shot 2019-03-06 at 08 22 09

In the example config they rely on less specific queries: 

screen shot 2019-03-06 at 08 20 47

*They* create the initial config for us so it'll be interesting to see what they do. I'm sure we'll have some back and forth and if they want unique ids on things we can come up with a convention / helper.

packages/documentation/src/components/seo.js
jbalboni: I don't think so?

PR #13, opened 2019-03-06
Add MDX document files for current formations React Components #16673

accbjt:
@jbalboni @rianfowler I have created all the documentation for the React components in formation-react. I moved all nested components outside into the main folder and refactored any import statement so it reflects these changes. I had to do a little refactoring of the MegaMenu so it can work with Gatsby. I also did some refactoring to the Gatsby layout file from some finding I found while doing the MDX pages.

The only problem I am having is that some of the css is not registering in the MegaMenu. It doesn't look correct. I am going to just make a new ticket (https://github.com/department-of-veterans-affairs/vets.gov-team/issues/17264) for that so we can get this PR in and merged.

I ran all the tests and also did `yalc` on vets-website repo to make sure nothing is broken. Please make sure you test this also, just in case I missed anything.

Ticket:
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/16673


packages/documentation/src/layouts/module-components.js
```diff
@@ -15,13 +15,26 @@ const MyCodeComponent = ({ children }) => (
 );
 
 export default class MDXRuntimeTest extends Component {
+  sortedReqiuredProps() {
```
accbjt: Added this so that we can have the required props on top of the list of props in the documentation.

packages/documentation/src/layouts/module-components.js
```diff
@@ -15,13 +15,26 @@ const MyCodeComponent = ({ children }) => (
...
+    );
+  }
+
   render() {
     const { children, data, tableOfContents } = this.props;
     // Window is not available during build, so this was added deal with making
     // window available to some components.
     // Tried to add it to componentDidMount but it didn't work correctly.
     if (typeof window !== 'undefined') {
       window.dataLayer = [];
```
accbjt: Window is not available on build so this is the workaround

packages/documentation/src/layouts/module-components.js
```diff
@@ -30,13 +43,27 @@ export default class MDXRuntimeTest extends Component {
           <div className="content">
             {children}
             <h2>{data.componentMetadata.displayName}</h2>
-            <p>{data.componentMetadata.docblock}</p>
+            {
```
accbjt: This makes the JsDoc block more readable in the docs. Before this is was 1 line without line breaks.

packages/documentation/src/scss/main.scss
```diff
@@ -79,3 +79,10 @@
     padding: 10px;
   }
 }
+
+.loading-indicator-container {
+  .loading-indicator {
```
accbjt: This is an override so the loader works in Gatsby

packages/documentation/src/pages/about-mdx.mdx
```diff
@@ -0,0 +1,105 @@
...
+
+### What is MDX?
+We will be using the .mdx format to create our documents.
+MDX is a format that lets you seamlessly use JSX in your Markdown documents.
+You can import components, like interactive charts or notifs, and export metadata.
+This makes writing long-form content with components a blast 🚀
+
+[https://mdxjs.com/](https://mdxjs.com/)
+
+### Example
```
rianfowler: `Examples`

packages/documentation/src/pages/about-mdx.mdx
```diff
@@ -0,0 +1,105 @@
...
+Here is a great page of examples of using MDX
+
+- http://dump.jamesknelson.com/mdxc-playground.html
+
+### Findings on things you can do with MDX and gotchas
```
rianfowler: Could simplify this to `Things you can do with MDX and gotchas`

packages/documentation/src/pages/about-mdx.mdx
```diff
@@ -0,0 +1,105 @@
...
+This is just documentations on finding of using MDX in our projects. Please feel
```
rianfowler: `This is documentation of findings for using...`

packages/documentation/src/pages/about-mdx.mdx
```diff
@@ -0,0 +1,105 @@
...
+free to add to this document if you find anything else.
+
+#### Inline components
+You can create components incline inside your MDX files. Below is an example
```
rianfowler: `You can use components inline inside` ... `files; Below`

`use` might be better as `demonstrate` or whatever activity we're intending them to do with this helper. Something like "you can demonstrate usage of your component" or something like that. I'm not sure exactly what this is for.

rianfowler: This first part might be good as a bulleted list 
```
- You can create ...
- You will need to export ...
- You can write both ...
```

packages/documentation/src/pages/about-mdx.mdx
```diff
@@ -0,0 +1,105 @@
...
+of how to do it. You will need to export it for it to work. You can write both
+function based components and class based components. The reason why you might
+use this is if your component updates the parent state. In the example you need
```
rianfowler: `In the example you need` -> maybe change the subject of this to the example: `This example uses state to toggle the drop down` unless we're directing them to implement this example?

packages/documentation/src/pages/about-mdx.mdx
```diff
@@ -0,0 +1,105 @@
...
+          </DropDownPanel>
+        </div>
+      </div>
+    )
+  }
+}
+
+<InlineClassComponent />
+
+// Function based component
```
rianfowler: `Function Component`

packages/documentation/src/pages/about-mdx.mdx
```diff
@@ -0,0 +1,105 @@
...
+
+#### Inline components
+You can create components incline inside your MDX files. Below is an example
+of how to do it. You will need to export it for it to work. You can write both
+function based components and class based components. The reason why you might
+use this is if your component updates the parent state. In the example you need
+state to toggle the drop down.
+
+```jsx
+// Classed Based Component
```
rianfowler: `Class Component`

packages/documentation/src/layouts/module-components.js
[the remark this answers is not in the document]
accbjt: @rianfowler Because this method gets directly invoked in the template it should already be bound to this, so I don't think we need to bind it. It works without needing to bind it.

packages/documentation/src/layouts/module-components.js
```diff
@@ -30,13 +43,27 @@ export default class MDXRuntimeTest extends Component {
...
+              data.componentMetadata.docblock && (
```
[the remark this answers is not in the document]
accbjt: I think for readability that is a good suggestion.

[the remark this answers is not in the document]
accbjt: The MDXRuntimeTest is used during the build process. It is connected through the gatsby system so I don't think we can actually do the handle the data changes before hand. It is defines as a layout template for the component pages.

PR #15, opened 2019-03-07
15360 us forms system when to use

rianfowler:
assumes mermaid support is incoming: 

screen shot 2019-03-07 at 12 36 06

screen shot 2019-03-07 at 11 55 11


packages/documentation/src/pages/when-to-use-usfs.mdx
```diff
@@ -0,0 +1,28 @@
+---
+title: When to use the US Forms System
+---
+
+## General structure of applications built with US Forms System
+
+The US Forms System provides a simple way to create a consistent form experience for veterans. Applications created with this framework follow this flow:
```
rianfowler: Not really sure how this document fits into the overall flow - we could expand the explanation to include more overview info or include this in another overview doc.

packages/documentation/src/pages/when-to-use-usfs.mdx
```diff
@@ -0,0 +1,28 @@
+---
+title: When to use the US Forms System
```
rianfowler: Is it appropriate to refer to this as the US Forms System?

packages/documentation/src/pages/when-to-use-usfs.mdx
```diff
@@ -0,0 +1,28 @@
+---
```
jbalboni: I think this file could go in the forms folder.

packages/documentation/src/pages/when-to-use-usfs.mdx
jbalboni: That's a good question. Maybe just "forms system"?

packages/documentation/src/pages/when-to-use-usfs.mdx
jbalboni: I think this probably would fit into an overview/introduction page.

packages/documentation/src/sidebar.js
```diff
@@ -74,6 +74,7 @@ module.exports = {
       items: [
         {
           name: 'Introduction',
+          href: 'forms/when-to-use-fs'
```
rianfowler: Just going to put this here for now. we can expand this document or reorganize it during an edit phase.

PR #16, opened 2019-03-08
Gatsby-remark-graph integration

accbjt:
@rianfowler I was trying to make this work with .mdx files but I could not get it to behave correctly. I was able to make it work with just .md files. This PR will allow us to write .md files in our project also and use the mermaid code snippet for graphs. I added some documentation on it and the documentation is actually an example of how to do it. If this looks good you can merge it in. I had to change some setting in the gatsby-node file to create pages from .md file located in the pages directory. Everything works pretty good, so far.

I also took a look at the gatsby-remark-graph and saw how they implemented it. They get the markdown AST and then convert it to <div> with a mermaid class. Then they have some code to inject the mermaid script and initialize it on the actual DOM element. Which converts it into the graph.

### DIFF
--- a/packages/documentation/gatsby-config.js
+++ b/packages/documentation/gatsby-config.js
@@ -63,6 +63,20 @@ module.exports = {
       }
     },
     `gatsby-transformer-react-docgen`,
-    `gatsby-transformer-remark`,
+    {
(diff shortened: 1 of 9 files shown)


jbalboni:
@accbjt Can we close this PR?

PR #19, opened 2019-03-08
create contextual hub variable names

jaredcunha-usds:
This PR will create a set of Sass variables for Hub colors that align with the naming conventions of all other colors in the design system. With this addition, front end developers will be able to use `color: $color-hub-healthcare` instead of `color: $indigo-cool-60`.

### DIFF
--- a/packages/formation/sass/base/_b-variables.scss
+++ b/packages/formation/sass/base/_b-variables.scss
@@ -90,6 +90,19 @@ $green-warm-50: #6f7a41;
 $blue-50: #357ab2;
 $violet-warm-60: #864381;
 
+// Contextual variable names
+$color-hub-health-care: $indigo-cool-60;
+$color-hub-education: $mint-cool-50;
+$color-hub-disability: $red-60;
+$color-hub-careers: $orange-warm-50;
+$color-hub-pension: $green-cool-50;
+$color-hub-housing: $gold-50;
+$color-hub-life-insurance: $green-warm-50;
+$color-hub-burials: $blue-50;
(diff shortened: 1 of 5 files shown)


packages/formation/sass/base/_b-variables.scss
```diff
@@ -90,6 +90,17 @@ $green-warm-50: #6f7a41;
...
+// Contextual variable names
+$color-hub-healthcare: $indigo-cool-60;
+$color-hub-education: $mint-cool-50;
+$color-hub-disability: $red-60;
+$color-hub-careers: $orange-warm-50;
+$color-hub-pension: $green-cool-50;
+$color-hub-housing: $gold-50;
+$color-hub-life-insurance: $green-warm-50;
+$color-hub-burials: $blue-50;
+$color-hub-records: $violet-warm-60;
```
jbalboni: I think we just added some more hubs in vets-website (see src/platform/site-wide/sass/iconography.scss). Should we add variables for those, too?

jaredcunha-usds: Oh, did not know. Yes, we should! I see they are the same color, so I'm just going to check with the designers to confirm

jaredcunha-usds: Added new hubs.

jaredcunha-usds:
Actually, I'm going to update the utilities to reflect the new Sass variables as well.

jaredcunha-usds:
@jbalboni I've updated the variables used in the utilities, so they now reference the new contextual names. Any other thoughts/question on this one?

PR #21, opened 2019-03-11
17320 mega menu ie 11 broken

rianfowler:
verified locally

### DIFF
--- a/packages/formation-react/src/components/MegaMenu/MegaMenu.jsx
+++ b/packages/formation-react/src/components/MegaMenu/MegaMenu.jsx
@@ -159,7 +159,7 @@ export default class MegaMenu extends React.Component {
                         {
                           item.title === currentDropdown && item.menuSections && <ul aria-label={item.title}>
                             {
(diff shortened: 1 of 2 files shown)


packages/formation-react/src/components/MegaMenu/MegaMenu.jsx
```diff
@@ -159,7 +159,7 @@ export default class MegaMenu extends React.Component {
                         {
                           item.title === currentDropdown && item.menuSections && <ul aria-label={item.title}>
                             {
-                              item.menuSections.constructor.name === 'Array' ? item.menuSections.map((section, j) => {
+                              Array.isArray(item.menuSections) ? item.menuSections.map((section, j) => {
```
rianfowler: this check fails on IE 11. 

sometimes `menusections` is an array. sometimes it's an object. its rendering behavior probably should be based on something more explicit than its datatype. 

Screen Shot 2019-03-11 at 15 16 52
---
Screen Shot 2019-03-11 at 15 16 42

rianfowler: Looks like this could have been broken for 7 months 😱

PR #23, opened 2019-03-12
Fix AdditionalInfo to handle aria-controls correctly - #1440

mjlumetta:
## Overview

PR for ticket 1440: https://github.com/department-of-veterans-affairs/vets-contrib/issues/1440

* Puts the `expandedContentId` of the `AdditionalInfo` component directly on the expandable content rendered as the second child of `ExpandingGroup` rather than as the `expandedContentId` prop for `ExpandingGroup` to align with expected `aria-controls` use
* Renders the expandable content within a hidden div in `ExpandingGroup` when the expandable content is collapsed so that the above change to `AdditionalInfo` does not break when the `aria-controls` element of the button trigger is absent
* Updates tests for these changes

## Testing

- [ ] Run the tests in `veteran-facing-services-tools`
- [ ] Run vets-website locally with the changes and confirm that they work on the /account page

### Running `vets-website` locally with changes

See the instructions here: https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/master/previewing-changes.md

Verify the following changes on the "What is ID.me" button:

* The button's `aria-controls` should still be the generated element id.
* When the section is collapsed (default), there should be a div with class `collapsed-content` and the `hidden` attribute set (as well as `aria-hidden="true"`). The collapsed content should be within this div with the generated id and the class of `additional-info-content`.

WhatIsIdmeCollapsed

* When the section is expanded, there should be a div with the class `expanded-content`. The expanded content should be in this div with the generate id and the class of `additional-info-content`.

WhatIsIdmeExpanded

* On enter, the expandable section should have CSS transitions expanding down and going from opacity 0 => 1.

### DIFF
--- a/packages/formation-react/package.json
+++ b/packages/formation-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "4.1.1",
+  "version": "4.1.2",
   "description": "React implementation of patterns in Formation",
   "keywords": [
     "react",
--- a/packages/formation-react/src/components/AdditionalInfo/AdditionalInfo.jsx
+++ b/packages/formation-react/src/components/AdditionalInfo/AdditionalInfo.jsx
@@ -7,7 +7,7 @@ import ExpandingGroup from '../ExpandingGroup/ExpandingGroup';
 export default class AdditionalInfo extends React.Component {
   constructor(props) {
     super(props);
-    this.expandedContentId = _.uniqueId('tooltip-');
+    this.expandedContentId = props.expandedContentId || _.uniqueId('tooltip-');
     this.state = { open: false };
   }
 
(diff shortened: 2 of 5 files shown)


1Copenut:
@jwoodman5 @mjlumetta 
I'm looking at the PR for this issue, and did not realize the outer `<span>` was being generated by Formation. I forgot to verify in JAWS that it worked as expected before writing [vets-contrib issue 1440](https://github.com/department-of-veterans-affairs/vets-contrib/issues/1440).

I have verified the existing code in JAWS + IE11, and it throws no automated errors, and the `aria-controls` works as expected.

In further digging, I also discovered tonight this pattern is being used on Markdown files, such as this one for [applying for healthcare](http://localhost:3001/health-care/how-to-apply/). In that MD file, the ID is also on the `<span>`, so changing the Formation instance would cause the two to fall out of sync.

I think keeping the code as is, and closing this issue and PR might be the best approach, with this new finding. I apologize for not catching this sooner, and leading to extra work.

PR #24, opened 2019-03-13
Add docs for common tasks and fill out section

jbalboni:
This adds our common tasks outlines and also adds docs for the simplest of those tasks. I'm going to create tickets for the rest of them, because they're a little more involved.

Screen Shot 2019-03-13 at 3 20 20 PM


packages/documentation/gatsby-node.js
```diff
@@ -97,93 +97,89 @@ exports.onCreateNode = ({node, getNode, actions }) => {
 
 exports.createPages = ({ graphql, actions }) => {
   const { createPage } = actions
-  return new Promise((resolve, reject) => {
-    resolve(
```
rianfowler: What does removing this call do?

jbalboni: The extra promise wrapping is unnecessary, so this is simplifying the code without changing any behavior. I changed it as part of my troubleshooting a build problem and left it in because it's an improvement either way.

PR #25, opened 2019-03-14
Add support for diagramming to documentation website Feautre/17294

accbjt:
Adding support for graphs for MDX files.


packages/documentation/gatsby-browser.js
```diff
@@ -0,0 +1,14 @@
+exports.onInitialClientRender = (n, config) => {
+  // eslint-disable-next-line
+  window.mermaid_config = { theme: config.theme || 'default', startOnLoad: true }
+  // load initial script
+  const s = document.createElement('script');
+  s.setAttribute('src', 'https://unpkg.com/mermaid@7.1.0/dist/mermaid.min.js');
+  document.head.appendChild(s);
+  // XXX: ugly hack because onRouteUpdate doesn't know when the react is done
+  setInterval(() => {
```
rianfowler: We might be able to use some lifecycle methods in [this](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/8efd9de68d277c1218eba52d289839c7e5786c1d/packages/documentation/src/layouts/module-components.js#L17) component to dispatch a custom event on `componentDidUpdate` and `componentDidMount`. 

We'd need to know when [this](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/8efd9de68d277c1218eba52d289839c7e5786c1d/packages/documentation/src/layouts/module-components.js#L65) component changes but it looks like it's going to update whenever `mdxCodeBlock` changes which is observable at this layer. My understanding is that lifecycle calls bubble from the inside out i.e. `child.componentDidMount` is called before `parent.componentDidMount`.

packages/documentation/plugins/remark/gatsby-remark-mdx-mermaid/index.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable */
+const visit = require('unist-util-visit');
+
+module.exports = ({ markdownAST }, { language = 'mermaid' } = {}) => {
+  visit(markdownAST, 'code', node => {
+    const lang = (node.lang || '').toLowerCase();
+
+    if (lang === language) {
+      node.type = 'html';
+      node.value = '<div class="mermaid">{' + '`'+ node.value + '`'+ '}</div>';
```
rianfowler: This opens up some possibilities- I wonder if we could use mermaid in node to generate the svg markup from this `node.value` and just inline it here. That would remove the need to run the client side library. 

I'm not sure exactly what's being created in this context: just markup or React code. We can generate React friendly svgs if needed (we did this on the old veteran id project- some of the xml directives had to be excluded but otherwise it worked fine).

packages/documentation/src/pages/getting-started/add-graphs-to-mdx.mdx
```diff
@@ -0,0 +1,33 @@
+---
+title: Add Graphs to MDX
+---
+
+## Add Graphs to MDX
+
+### Mermaid Library
+To generator graphs in MDX files, we need to use the [mermaid](https://mermaidjs.github.io/) library.
```
rianfowler: `To generate`

packages/documentation/plugins/remark/gatsby-remark-mdx-mermaid/index.js
accbjt: @rianfowler I actually tried to do that but the only way to make it work with node is to use the mermaidCLI. I have't tried to implement the CLI yet but if we were to try to do it we would need to generate a .mmd file and then convert that into a svg and then add that link to our html document. Mermaid is client side and you won't be able to use it unless you have the window object and also the SVG object. I even tried it with the mermaidAPI using jsdom but that didn't have everything we needed to make it work. It would bomb out when trying to generate the SVG. I would be open to trying the CLI but I think that would take a little time to implement.

packages/documentation/gatsby-browser.js
accbjt: We can definitely look into this more.

jbalboni: Yeah, I think we should see if we can import mermaid in layout.js and use componentDidMount in that component to do the initialization.

packages/documentation/plugins/remark/gatsby-remark-mdx-mermaid/index.js
rianfowler: Thanks for the info- that's really frustrating that it doesn't just generate the markup and requires DOM stuff. 

this approach is good- if we want we can try the CLI later. I tried the MMD file approach but I think the trade off isn't worth it unless we have a lot of diagrams. Would require a bunch of file structure or naming conventions that could be annoying.

packages/documentation/src/pages/getting-started/create-gatsby-remark-plugin.mdx
```diff
@@ -0,0 +1,131 @@
...
+    A --> C(Round Rect)
+    B --> D{Rhombus}
+    C --> D
+```
+
+Remark uses a plugin system that enables you to do some customization when the markdown gets converted into html through Remark. That is what we are going to use to convert our markdown into a graph.
+
+- First we need to create a plugin that we can add to the `gatsby-transformer-remark` plugin.
+
+  Below is the code we are using to generate the html to convert our code into graphs. When adding this to the plugin, this code will run during the transformation phase. In this code we are taking the `mardownAST` and finding mermaid specific code on the page. Then we change the code to html.
```
rianfowler: `markdownAST`

packages/documentation/plugins/mermaid-svg-remover/gatsby-node.js
```diff
@@ -0,0 +1,27 @@
+const fs = require('fs');
+const path = require('path');
+
```
accbjt: This will remove all SVG files generated by the mermaid.cli. We have no way to track if a mermaid graph is removed from a document. This will reset it so it will only have graphs that are on current pages.

packages/documentation/plugins/remark/gatsby-remark-mdx-mermaid/index.js
```diff
@@ -0,0 +1,27 @@
...
+const remark = require('remark');
+const mermaid = require('./remark-mermaid');
+const vfile = require('to-vfile');
+
+module.exports = ({ markdownAST, markdownNode }, { language = 'mermaid', theme = 'default' } = {}) => {
+  visit(markdownAST, 'code', async (node) => {
+    const lang = (node.lang || '').toLowerCase();
+    const fileContent = vfile.readSync(markdownNode.fileAbsolutePath);
+
+    if (lang === language) {
```
accbjt: I had to use remark directly to make it work with the mermaid.cli.

packages/documentation/plugins/remark/gatsby-remark-mdx-mermaid/remark-mermaid/src/index.js
```diff
@@ -0,0 +1,207 @@
+const fs = require('fs-extra');
```
accbjt: I had to modify the original version of mermaid to make this work with Gatsby. The original program didn't give us back the link and I needed to make sure the SVGs get generated in our static folder so we can use it on our pages.

packages/documentation/plugins/remark/gatsby-remark-mdx-mermaid/remark-mermaid/src/index.js
```diff
@@ -0,0 +1,207 @@
...
+      vFile.info(`${lang} code block replaced with div`, position, PLUGIN_NAME);
+
+    // Otherwise, let's try and generate a graph!
+    } else {
+      let graphSvgFilename;
+
+      try {
+        graphSvgFilename = render(value, destinationDir);
+
+        vFile.info(graphSvgFilename, position, PLUGIN_NAME);
```
accbjt: This is where I return a object with the specific code line and path to the svg image.

packages/documentation/plugins/remark/gatsby-remark-mdx-mermaid/remark-mermaid/src/utils.js
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Accepts the `source` of the graph as a string, and render an SVG using
+ * mermaid.cli. Returns the path to the rendered SVG.
+ *
+ * @param  {string} source
+ * @param  {string} destination
+ * @return {string}
+ */
+function render(source, destination) {
+  const staticFilePath = `${process.cwd()}/static/mermaid-svgs`;
```
accbjt: changed the path to use our static folder

packages/documentation/plugins/remark/gatsby-remark-mdx-mermaid/remark-mermaid/src/utils.js
```diff
@@ -0,0 +1,92 @@
...
+  // Write temporary file
+  fs.outputFileSync(mmdPath, source);
+
+  // Invoke mermaid.cli
+  execSync(`${mmdcExecutable} -i ${mmdPath} -o ${svgPath} -b transparent`);
+
+  // Clean up temporary file
+  fs.removeSync(mmdPath);
+
+  return { source, imgSrc: `/mermaid-svgs/${svgFilename}`};
```
accbjt: this returns the path to the image we will use on our page. Source gives us back an object regarding what mermaid has changed.

packages/documentation/plugins/remark/gatsby-remark-mdx-mermaid/index.js
```diff
@@ -0,0 +1,27 @@
...
+      await remark()
+        .use(mermaid)
+        .process(fileContent, (err, file) => {
+          if (err) throw err;
+
+          const mermaidSrcMatch = file.messages.filter(
```
accbjt: finds the specific mermaid code line that matches the current SVG. This handles the case where there are multiple mermaid graphs on a page.

packages/documentation/package.json
```diff
@@ -8,6 +8,7 @@
     "@mdx-js/mdx": "^0.16.6",
     "@mdx-js/tag": "^0.16.6",
     "classnames": "^2.2.6",
+    "eslint-plugin-react": "^7.12.4",
```
jbalboni: Why is this added?

jbalboni:
@accbjt It looks like you deleted yarn.lock and added package-lock.json. Was that intentional? I think we should keep yarn.lock and not have a package-lock.json file.

packages/documentation/package.json
accbjt: Not sure why this got added

accbjt:
@jbalboni Yeah I actually didn't want the package-lock file. I always use yarn. I'm not sure how this got into the project. It happened when I pulled from master. Some how this got added. It looks like @jaredcunha-usds added this by accident. Here's the commit it got added to master. @jbalboni do you want me to remove the package-lock file from the project or do you want to do it in another branch?

https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/commit/0c8013472cddd1e9b1f5dfd3eeb6ae01ddd8a4e4#diff-32607347f8126e6534ebc7ebaec4853d

jbalboni:
Yeah, lets remove it and restore yarn.lock

accbjt:
Okay I will do that.

packages/documentation/plugins/mermaid-svg-remover/gatsby-node.js
```diff
@@ -0,0 +1,27 @@
...
+exports.onPreBootstrap = async () => {
+  const staticDir = path.join(__dirname, '../../static/mermaid-svgs');
+  const publicDir = path.join(__dirname, '../../public/mermaid-svgs');
+
+  await fs.readdir(staticDir, (err, files) => {
+    if (err) throw err;
+
+    for (const file of files) {
+      fs.unlink(path.join(staticDir, file), err => {
```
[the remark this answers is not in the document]
accbjt: Good catch. I will fix it.

packages/documentation/static/mermaid-svgs/128cea220b14b53b1de9cd7644a533da7ae1fca5.svg
```diff
@@ -0,0 +1,4 @@
...00!important}#mermaid-1552891724961 .titleText{text-anchor:middle;font-size:18px;fill:#000}#mermaid-1552891724961 g.classGroup text{fill:#9370db;stroke:none;font-family:trebuchet ms,verdana,arial;font-size:10px}#mermaid-1552891724961 g.classGroup rect{fill:#ececff;stroke:#9370db}#mermaid-1552891724961 g.classGroup line{stroke:#9370db;stroke-width:1}#mermaid-1552891724961 .classLabel .box{stroke:none;stroke-width:0;fill:#ececff;opacity:.5}#mermaid-1552891724961 .classLabel .label{fill:#9370db;font-size:10px}#mermaid-1552891724961 .relation{stroke:#9370db;stroke-width:1;fill:none}#mermaid-1552891724961 #compositionEnd,#mermaid-1552891724961 #compositionStart{fill:#9370db;stroke:#9370db;stroke-width:1}#mermaid-1552891724961 #aggregationEnd,#mermaid-1552891724961 #aggregationStart{fill:#ececff;stroke:#9370db;stroke-width:1}#mermaid-1552891724961 #dependencyEnd,#mermaid-1552891724961 #dependencyStart,#mermaid-1552891724961 #extensionEnd,#mermaid-1552891724961 #extensionStart{fill:#9370db;stroke:#9370db;stroke-width:1}#mermaid-1552891724961 .branch-label,#mermaid-1552891724961 .commit-id,#mermaid-1552891724961 .commit-msg{fill:#d3d3d3;color:#d3d3d3}</style><style>#mermaid-1552891724961 {
```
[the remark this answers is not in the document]
rianfowler: Do these added to the `gh-pages` branch during the build / publish process?

rianfowler:
The way I understand this, we can use the `mermaid` syntax in the markdown files and this will 
- generate the svg
- remove the mermaid code block, and
- add a link to the svg in its place.

jaredcunha-usds:
Whoops. Sorry about that!

accbjt:
@rianfowler Yes that is correct.

packages/documentation/plugins/mermaid-svg-remover/gatsby-node.js
```diff
@@ -0,0 +1,27 @@
...
+
+exports.onPreBootstrap = async () => {
+  const staticDir = path.join(__dirname, '../../static/mermaid-svgs');
+  const publicDir = path.join(__dirname, '../../public/mermaid-svgs');
+
+  await fs.readdir(staticDir, async (err, files) => {
+    if (err) throw err;
+
+    for (const file of files) {
+      await fs.unlink(path.join(staticDir, file), err => {
```
jbalboni: Does this work? I don't think this method returns a promise, does it? I guess I have similar questions about the rest of this code, too.

packages/documentation/plugins/remark/gatsby-remark-mdx-mermaid/index.js
```diff
@@ -0,0 +1,27 @@
...
+            item => item.message.source === node.value
+          )[0];
+
+          node.type = 'html';
+          node.value = `<img src="${mermaidSrcMatch.message.imgSrc}" />`;
```
jbalboni: I'm going to guess that this will break on the github.io site, since it won't have the site prefix in this path. I'm not sure how easy that is to get in here.

packages/documentation/plugins/mermaid-svg-remover/gatsby-node.js
accbjt: @jbalboni Yeah your right this should return a promise. Wow I'm rusty with Node.

PR #26, opened 2019-03-14
add prepare command

rianfowler:


packages/formation/package.json
```diff
@@ -18,7 +18,8 @@
     "url": "git+https://github.com/department-of-veterans-affairs/veteran-facing-services-tools.git"
   },
   "scripts": {
-    "build": "node ./scripts/build.js"
+    "build": "node ./scripts/build.js",
+    "prepare": "node ./scripts/build.js"
```
rianfowler: `lerna bootstrap` symlinks local peer packages instead of installing them from `npm`. `prepare` is run by `lerna bootstrap` on each package- this prompts `formation` to build it's scss properly for use in the `documentation` package as a symlink during bootstrapping.

PR #27, opened 2019-03-15
add missing uswds variables

jaredcunha-usds:
This PR will add some USWDS variables to Formation. 

With these additions, you can use the Formation Sass variables and functions without requiring USWDS as a dependency.

### DIFF
--- a/packages/formation/sass/base/_b-breakpoints.scss
+++ b/packages/formation/sass/base/_b-breakpoints.scss
@@ -13,6 +13,16 @@
 // Based on the order of our imports, USWDS should override Foundation's declarations and everything should be fine.
 // However, USWDS defines its variables using the "!default" flag, so Foundation's take precedence anyway.
 // These next lines redeclare USWDS's variables without that flag, so that the override works as we would expect.
+
+$grid-columns-small: 1 !default;
+$grid-columns-medium: 6 !default;
+$grid-columns-large: 12 !default;
+
+// Breakpoint variables from USWDS 1.4.2
+$small-screen:  481px !default;
+$medium-screen: 600px !default;
+$large-screen:  1201px !default;
+
 $small: new-breakpoint(min-width $small-screen $grid-columns-small);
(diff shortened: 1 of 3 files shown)


packages/formation/sass/base/_b-breakpoints.scss
```diff
@@ -13,6 +13,16 @@
 // Based on the order of our imports, USWDS should override Foundation's declarations and everything should be fine.
 // However, USWDS defines its variables using the "!default" flag, so Foundation's take precedence anyway.
 // These next lines redeclare USWDS's variables without that flag, so that the override works as we would expect.
+
+$grid-columns-small: 1 !default;
```
jaredcunha-usds: Reading the note above, did we get around to removing Foundation classes? I kept the `!default` flags in there, but wasn't sure.

packages/formation/sass/base/_b-variables.scss
```diff
@@ -12,9 +12,16 @@
 // ~uswds/src/stylesheets/core/defaults" or
```
jaredcunha-usds: Just a note: any of the lines indicated to be removed were moved in more logical groupings.

packages/formation/sass/base/_b-breakpoints.scss
jbalboni: We've removed the visibility imports, I think, but the grid is a bigger project that we haven't gotten to yet.

jaredcunha-usds: Is there any way you can confirm that these won't overwrite the values in vets-website?  

I assume because of the !default flag, it should be safe because it's supposed to just set the variable if it's not already set.

jbalboni: Yeah, I believe that's how it works. It's not used in vets-website that I can see, so it should be safe. I think to test it specifically we'd need to make a class that used it and verify that the output was correct.

PR #28, opened 2019-03-19
Feature/14003-JAWS-ARIA

1Copenut:
## PR Purpose
The JAWS screen reader was not working correctly with our Mega menu. I observed the screen reader opening the menu, but the virtual cursor disappearing behind the menu, and not moving into the menu as expected. 

After careful analysis, I determined the `aria-haspopup="true"` attribute was causing JAWS to move from **Browse** mode to **Forms** mode. Forms mode is a circuit-breaker for the normal arrow key behavior. Pressing `DOWN_ARROW`  no longer moves the virtual cursor to the next DOM element. It is meant to handle elements like radio buttons and select menus, so is not what we want in this scenario.

Removing the `aria-haspopup` attribute allows JAWS to stay in Browse mode, and the virtual cursor works as expected.

## Related Issue
* https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/14003

## Tested
* VoiceOver + Safari - MacOS
* NVDA + Firefox - Win10
* JAWS + IE11 - Win10
* JAWS + Chrome - Win10

## Added
* New `role` attribute
* ID to containing divs so `aria-controls` will work with JAWS

## Removed
* Several `aria` and `role` attributes from the MegaMenu sub-components
* A few SCSS selectors to update styling appropriately


packages/formation/sass/modules/_m-megamenu.scss
```diff
@@ -90,7 +93,7 @@ $marketing-container-height: 380px;
       white-space: normal;
 
       h3 {
-        display: initial;
+        display: block;
```
1Copenut: This was changed so the headings would appear in IE11. They were continuing to hold onto the `display: none` declaration.

packages/formation/sass/site/_m-vet-nav.scss
```diff
@@ -216,7 +216,7 @@ body.va-pos-fixed {
     }
   }
 
-  &[aria-haspopup]:not([aria-expanded="true"]) {
+  &[aria-expanded='false'] {
```
1Copenut: Removing the `[aria-haspopup]` was why I opted to do major version bumps. I don't anticipate breaks because of higher selector specificity, but it's a low cost way of being eyes up.

packages/formation-react/src/components/MegaMenu/SubMenu.jsx
```diff
@@ -1,16 +1,16 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-import Column from './Column';
-import _ from 'lodash';
-import { ArrowRightBlueSVG } from './arrow-right-blue';
+import React from "react";
```
jbalboni: It looks like the wrong Prettier rules got applied in here. @1Copenut You might need to get the latest and run the lint commands again (cc @rianfowler )

rianfowler: [This](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/35#issue-263072113) should fix it once it's merged. Feel free to merge on approval.

1Copenut:
Re-tested on 3/22 after linting updates. The JAWS screen reader is moving to the mega menu correctly now as expected.

packages/formation-react/src/components/MegaMenu/SubMenu.jsx
```diff
@@ -21,6 +21,7 @@ const SubMenu = ({
   mobileMediaQuery,
   smallDesktopMediaQuery,
   data,
+  id,
```
rianfowler: I don't see where this id is coming from- was there another PR that added this?

1Copenut: This ID came from the `<MenuSection />` component. I passed it down and made it a named prop.

Previously it was being applied to the button. It was needed on the target element so JAWS would properly move to the controlled element when users entered the key chord `Alt + JAWS_Key + M`. 

Line 66 of MenuSection.jsx: https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/ebf5e2f531d9bc1a3cc307a5f73ec6700673b20f/packages/formation-react/src/components/MegaMenu/MenuSection.jsx#L66

1Copenut:
@rianfowler Before I merge this into master, wanted to get an opinion on the version bump. Do you feel these need to be major jumps to 5.0.0 and 6.0.0 respectively, or would they be better served as minor bumps to 4.2.0 on formation-react and 5.6.0 on formation?

I can go either way, and will continue testing in vets-website before issuing a PR to bump that codebase.

rianfowler:
minor should be adequate- it implies that the change is not backward compatible. major is usually reserved for significant upgrades or additions to feature sets.

1Copenut:
> minor should be adequate- it implies that the change is not backward compatible. major is usually reserved for significant upgrades or additions to feature sets.

I'll update the `package.json` numbers, make sure tests pass, and merge it.

packages/formation/sass/site/_m-vet-nav.scss
1Copenut: On further discussion, we are going to call this a `minor` bump. The Formation codebase doesn't use `aria-haspopup` anywhere else for CSS styling. Did a full-text search for the strings "aria-haspopup" and "haspopup" in the Formation code base locally, and came up with no results.

PR #29, opened 2019-03-19
Add http headers deck

rianfowler:
there's another ticket to add support for markdown slides but you can see a preview of the slides by pasting this into here: 

https://web.marp.app

if we decide to support slides / decks, interested to hear how we would like to structure them in the doc repo. we could duplicate the file, remove the slide specific markup, put a link to the deck in that file, and exclude the deck from the search indexing.


rianfowler:
I was also thinking could restructure the content around "what you can control with http headers" instead of just a list of all of our http headers.

PR #30, opened 2019-03-19
Getting started section

jbalboni:
This adds docs for the getting started section, along with some other orientation content.

There's not a lot here and we probably need to add more backend set up help. I've created a ticket to do that separately.


packages/documentation/yarn.lock
```diff
@@ -10009,11 +10009,6 @@ preserve@^0.2.0:
   resolved "https://registry.yarnpkg.com/preserve/-/preserve-0.2.0.tgz#815ed1f6ebc65926f865b310c0713bcb3315ce4b"
   integrity sha1-gV7R9uvGWSb4ZbMQwHE7yzMVzks=
 
-prettier@^1.15.2:
```
jbalboni: No idea how that happened, I'll remove the change.

PR #31, opened 2019-03-20
Refactor Document project to use locally created plugins Feature/17361

accbjt:
We need to now refactor the Documentation code so it can use the local plugin system that Gatsby has. Currently all the code we use are located in one file and is not very maintainable. We need to abstract this out to it's own plugin. We also have to document how to do.


packages/documentation/yarn.lock
```diff
@@ -10009,11 +10009,6 @@ preserve@^0.2.0:
   resolved "https://registry.yarnpkg.com/preserve/-/preserve-0.2.0.tgz#815ed1f6ebc65926f865b310c0713bcb3315ce4b"
   integrity sha1-gV7R9uvGWSb4ZbMQwHE7yzMVzks=
 
-prettier@^1.15.2:
```
jbalboni: Not sure why this keeps happening. Can you remove this change?

PR #36, opened 2019-03-21
Refactor GitHub API integration on the Documentation project so it uses JSON file Feature/17436

accbjt:
Currently our Documentation project, the GitHub API integration uses github directory names to import markdown files into the project. We need a better manageable way to deal with this. We should have some sort of JSON file that will manage what is imported into the project.


packages/documentation/package.json
```diff
@@ -31,7 +31,6 @@
     "react-dom": "^16.6.3",
     "react-helmet": "^5.2.0",
     "react-live": "^1.12.0",
-    "remark-mermaid": "^0.2.0",
```
accbjt: we don't need this anymore.

accbjt:
@jbalboni @rianfowler Let me know if this implementation looks okay.

packages/documentation/plugins/github-api-pages/gatsby-node.js
```diff
@@ -1,37 +1,18 @@
-const path = require('path');
+const path  = require('path');
```
jbalboni: Looks like this was accidental?

packages/documentation/src/pages/getting-started/common-tasks/creating-gatsby-plugins.mdx
```diff
@@ -0,0 +1,63 @@
+---
```
jbalboni: Can we add this page to the sidebar in common tasks?

packages/documentation/plugins/github-api-pages/pages/index.js
```diff
@@ -0,0 +1 @@
+exports.workPractices = require('./work-practices.json');
```
jbalboni: I think we could just have one JSON file, until we get to the point where that's too painful (which I think is unlikely).

packages/documentation/plugins/github-api-pages/gatsby-node.js
accbjt: okay

accbjt: Yeah I think it was

PR #38, opened 2019-03-21
add heading levels to font size utilities

jaredcunha-usds:
This will add some extra classes to the font-size utility. The classes are redundant, but @1Copenut mentioned it would help developers choose the right utility class when change font sizes on headings while maintaining correct semantic level.


jaredcunha-usds:
Does anybody know why the checks are failing? I can't view the details,

PR #39, opened 2019-03-21
ignore built assets in formation react; fix linting errors

rianfowler:
- add rule to ignore built resources in formation-react
- fix linting issues

### DIFF
--- a/packages/vagov-eslint/.eslintignore
+++ b/packages/vagov-eslint/.eslintignore
@@ -5,3 +5,4 @@
 **/.cache/
 **/*.config.*
 **/*.conf.*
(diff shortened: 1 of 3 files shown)


packages/vagov-eslint/.eslintignore
```diff
@@ -4,4 +4,5 @@
 **/node_modules/*
 **/.cache/
 **/*.config.*
-**/*.conf.*
\ No newline at end of file
+**/*.conf.*
+/packages/formation-react/*.js
```
rianfowler: might revisit this ignore file later and require each package to maintain it's own .eslintignore

PR #41, opened 2019-03-21
17368 absolute image paths

rianfowler:
- updates webpack config to create output directories using `file-loader`
  - `./dist/fonts` receives all picked up by the font test
  - `./dist/img` receives all files not inlined by `svg-url-loader` and `url-loader`
- removes directory creation from build script 
- adds `/dist` cleanup to build script 
- points `background-image` references in scss to correct files- `file-loader` ignores any images it cannot find

Screen Shot 2019-03-21 at 16 48 14
Screen Shot 2019-03-21 at 16 47 54


packages/formation/sass/modules/_m-loading-indicator.scss
```diff
@@ -3,7 +3,7 @@
 
   .loading-indicator {
     min-height: 8em;
-    background: url(/img/loading-state.svg) center center no-repeat;
+    background: url(../assets/img/loading-state.svg) center center no-repeat;
```
rianfowler: `url-loader` and `svg-url-loader` look for these files relative to the entry file (`./sass/full.scss`) and does nothing when they cannot find them. This updates the references so these loaders copies them into the directories according to the rules in the webpack config.

packages/formation/scripts/build.js
```diff
@@ -1,13 +1,12 @@
...
 const webpack = require('webpack');
 const webpackConfig = require('../../../webpack.config.prod.js');
+const rimraf = require('rimraf');
 
 console.log('Starting Formation build');
 console.log('Cleaning old build');
-mkdirp.sync('./dist/img');
-mkdirp.sync('./dist/fonts');
+
+rimraf.sync('./dist');
```
rianfowler: Updates build script to actually clean up the old build.

packages/formation/scripts/build.js
```diff
@@ -1,13 +1,12 @@
 /* eslint-disable no-console */
-const mkdirp = require('mkdirp');
-const fs = require('fs-extra');
 const webpack = require('webpack');
 const webpackConfig = require('../../../webpack.config.prod.js');
+const rimraf = require('rimraf');
 
 console.log('Starting Formation build');
 console.log('Cleaning old build');
-mkdirp.sync('./dist/img');
```
rianfowler: Webpack can do this. We could potentially ditch this build script and use webpack's CLI.

webpack.config.prod.js
```diff
@@ -29,17 +30,13 @@ module.exports = {
       {
         test: /\.(jpe?g|png|gif)$/i,
         use: {
-          loader: 'url-loader?limit=10000!img?progressive=true&-minimize',
-          options: {
-            name: '[name].[ext]',
-            outputPath: 'img/'
```
rianfowler: `url-loader` and `svg-url-loader` fallback to `file-loader` when a file exceeds the limit set by their respective options. `outputPath` is strictly a `file-loader` option but it will only receive this option when it's set as a query param. With the old configuration, `file-loader` would simply not work. 

Relying on `file-loader` instead of the build script to copy the files that are not inlined results in a cleaner dist directory that contains only the assets referenced in the output files.

webpack.config.prod.js
```diff
@@ -29,17 +30,13 @@ module.exports = {
...
-          }
+          loader: 'url-loader?limit=10000!img?progressive=true&-minimize&outputPath=./img/',
```
rianfowler: I tested reducing this limit but there are a lot of images that are inlined by imported `.usa` styles that we aren't impacted by this rule. We could potentially reduce the size of our CSS by finding a way to manage the inlining of these images.

webpack.config.prod.js
```diff
@@ -9,7 +9,8 @@ module.exports = {
   },
   output: {
     filename: '[name].js',
-    path: path.join(__dirname, 'packages/formation/dist')
+    path: path.join(__dirname, 'packages/formation/dist'),
+    publicPath: './'
```
rianfowler: We can use `publicPath` to set the output files to use a relative path for their urls.

webpack.config.prod.js
rianfowler: I also don't know what `!img` does.

jaredcunha-usds:
This what I see in my sandbox. As you can see, the images are working. 

Screen Shot 2019-03-22 at 9 26 58 AM

This what it looks like on VA.gov (CSS for some items are on `vets-website`)
Screen Shot 2019-03-22 at 9 33 17 AM


It does not appear that Bitter is getting included in the fonts directory. Here's the 5.5.2 tree that I used installing via `npm`
Screen Shot 2019-03-22 at 9 29 36 AM

Here's the same directory of this branch:
Screen Shot 2019-03-22 at 9 30 01 AM

jaredcunha-usds:
Then out of curiosity, what are these strange svgs that were removed?
Screen Shot 2019-03-22 at 9 40 50 AM

rianfowler:
@jaredcunha-usds 
- I'll see what's going on with the bitter fonts.
- this set up differs from what we had before 
  - before, we used `build.js` to manually copy _all_ of the assets over. most of the svgs were passed over by the `svg-url-loader` because they were referencing an `/img/name.svg` that webpack couldn't find in the `/src` but worked in `/dist` because ` build.js` copied them into the final output. since they were passed over, the `svg-url-loader` would never attempt to inline the svg into the css even if it was below the limit set in the config. 
  - this approach puts all of the `dist` creation under webpack so it can decide if images get copied into the `/dist` or inlined into the css. I can try to get it to preserve the original file name but I think the default behavior is to use a hashed file name (this shouldn't change between builds if the file is updated).

jaredcunha-usds:
@rianfowler - Is it a huge lift to have all the images copied over into the `img` folder with the original file name, even if they are embedded in the CSS? Thinking in case someone wants to use those images in a project file or on a local, non-Formation component.

If that's a huge lift, we can figure out how to tackle later. I don't think this would need to be a blocker.

rianfowler:
@jaredcunha-usds We can accommodate whatever is needed. We could generate multiple CSS, too- one that inlines all of the images to one that inlines none of the ones from formation for maximum flexibility. The inlining is mostly a very minor performance optimization.

I personally think that sites should only inline css images that are on landing pages regardless of the size- the `.usa` images is a good example of that. this CSS file is big (80kb) and there's not a lot of opportunity to make it smaller by removing inlined images from that external dependency the way it's set up currently. 

I'll look into preserving the original file names.

jaredcunha-usds:
I don't think we need multiple CSS files. I think we should inline the images, as you have done. Just wanted to make that all of the images can be placed into the `img` folder so they can be used in other project css selector, if someone wanted to do that.

packages/formation/package.json
```diff
@@ -27,7 +27,8 @@
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.6.3",
     "foundation-sites": "5",
-    "moment": "^2.23.0"
+    "moment": "^2.23.0",
```
jbalboni: These are maybe dev dependencies?

jbalboni:
Looks like we need to delete some images from that assets folder. The ones in the docs and design folder shouldn't be distributed with formation.

jbalboni:
@rianfowler Have you tried testing this with vets-website? I get a bunch of errors about not finding images now. I'm guessing we need some webpack config changes there to handle this.

packages/formation/.npmignore
```diff
@@ -1,2 +1,3 @@
 scripts/*
-assets/*
+assets/img/docs/*
+assets/img/design/*
```
rianfowler: updated ignores used during publishing

packages/formation-react/src/components/AdditionalInfo/AdditionalInfo.unit.spec.jsx
```diff
@@ -32,7 +32,7 @@ describe('<AdditionalInfo/>', () => {
     wrapper.unmount();
   });
   it('renders both children when open is true', () => {
-    const first = wrapper.find('ExpandingGroup').props();
+    const first = wrapper.find(ExpandingGroup).props();
```
rianfowler: I don't know why this started failing on my branch- suspect there might be a bigger issue with our unit tests.

packages/formation/sass/modules/_m-homepage-hero.scss
```diff
@@ -4,7 +4,6 @@
 
 .homepage-hero-image {
   align-items: center;
-  background: $color-white url(/img/design/background/veterans.jpg) center center no-repeat;
```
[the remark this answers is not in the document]
jbalboni: Yeah, I think this is old

jbalboni:
@rianfowler Our tests run in Karma, which has a webpack plugin and also an abbreviated config in karma.conf.js. Might need to be updated/tweaked with your changes. That could be the source of the testing issues.

packages/formation-react/testing/karma.conf.js
```diff
@@ -15,6 +15,7 @@ module.exports = function karmaConfig(config) {
     frameworks: ['mocha'],
     webpack: { // kind of a copy of your webpack config
       devtool: 'inline-source-map', // just do inline source maps instead of the default
+      mode: 'none',
```
rianfowler: some of our enzyme tests rely on querying a component's `displayName` using `node.find()`. `displayName` is only included in non production builds and webpack 4 defaults to production builds if no `mode` is set. `none` forces webpack to previous versions default behavior. 

Not exactly where what's doing this optimization- webpack or babel or react.

packages/formation/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "5.6.1",
+  "version": "5.7.0",
```
rianfowler: formation requires scss variable overrides

rianfowler:
@jaredcunha-usds not sure if this is in the design documentation, but we should have a blurb similar to the one in [uswds](https://github.com/uswds/uswds#set-the-base-asset-path-fonts-and-images). I named the asset paths `$formation-asset-path`, `formation-image-path` and `formation-font-path`. If developers import any of the scss directly, they will need to set these in their root scss file.

packages/formation/.npmignore
jbalboni: Could we delete them from the repo, too?

packages/formation/package.json
jbalboni: I think this needs to go up a major version. You have to update vets-website in order to build styles using these changes.

jbalboni:
@rianfowler @jaredcunha-usds It looks like the formation.min.css is looking for relative paths now, but it's looking in `./assets/img` or `./assets/fonts`. Are we ok with that, or do we want to remove the assets part of that path? Not sure if that's possible or how easy it is, but might be simpler for consumers of the built css.

jaredcunha-usds:
@jbalboni 
>@rianfowler @jaredcunha-usds It looks like the formation.min.css is looking for relative paths now, but it's looking in ./assets/img or ./assets/fonts. Are we ok with that, or do we want to remove the assets part of that path? Not sure if that's possible or how easy it is, but might be simpler for consumers of the built css.

I would prefer to remove the assets part if that's doable, but if someone can do something like this wouldn't be the end of the world: 
```
project_root
   |----my_assets_directory
               |----formation
                               |---- assets/*
                               |---- formation.min.css
```

@rianfowler 
> @jaredcunha-usds not sure if this is in the design documentation, but we should have a blurb similar to the one in uswds. I named the asset paths $formation-asset-path, $formation-image-path and $formation-font-path (i can update these if you want). If developers import any of the scss directly, they will need to set these in their root scss file.

Shoud we use the `!default` flag on those asset path declarations so they can override in their project files and safely update Formation?

rianfowler:
@jaredcunha-usds yeah- the dist directory is pretty straight forward to changes. LMK if this works 
[formation.zip](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/files/3034524/formation.zip)

jaredcunha-usds:
@rianfowler Love it!

PR #43, opened 2019-03-22
Remove greater than errorable radio mg

markgreenburg:
Found a bug where an extra `>` was getting prepended to labels in radio buttons.

Before:
Screen Shot 2019-03-22 at 9 39 06 AM

After:
Screen Shot 2019-03-22 at 9 22 32 AM

### DIFF
--- a/packages/formation-react/package.json
+++ b/packages/formation-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "4.1.2",
+  "version": "4.1.3",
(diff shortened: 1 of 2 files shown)


packages/formation-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "4.1.2",
+  "version": "4.1.3",
```
ncksllvn: I think we also need to update the `yarn.lock`

PR #44, opened 2019-03-25
Remove autocomplete attribute from checkboxes, radio buttons, and sel…

mjlumetta:
…ects

PR for [vets-contrib#1609](https://github.com/department-of-veterans-affairs/vets-contrib/issues/1609). Completes some of the work scoped in [this ticket](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15045).

Removes autocomplete from `input` elements with type `checkbox`, `radio`, and `number` and `select` elements. This will allow components to pass checks using `axe-core` with version > 3.0.

### DIFF
--- a/packages/formation-react/package.json
+++ b/packages/formation-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "4.2.0",
+  "version": "4.2.1",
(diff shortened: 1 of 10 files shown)


jbalboni:
@dvogel Once we get axe-core updated, hopefully this week, we should catch any future regressions.

PR #45, opened 2019-03-27
Megamenu 17558

ncksllvn:
The addition of two new hubs with no child pages to VA.gov has introduced a new case for the MegaMenu - we should be able to add items into the sidebar of the "VA Benefits and Health Care" dropdown, but as plain links rather than section titles. 

I hope this isn't an issue, but I also used this opportunity to do some cleaning up of the MegaMenu component. I had somewhat of a hard time following the code and thought it could use some housekeeping instead of just searching for a CSS selector and adding more code. Ultimately, this PR -

- Chunks out a `<Navbar/>` component from the main `<MegaMenu/>` component. I
I noticed a couple other things I think we could start doing to improve have vet-nav SCSS still in vets-website that should be ported.
- Always renders the elements of the dropdown and dropdown submenus, and instead just toggles their visibility. I believe this will help SEO a bit by allowing all links to always exist in the DOM rather than rendered only on-demand, and possibly make the experience smoother for screen readers. It also makes the menu a bit snappier since the complex rendering is done only on the first render - there's a noticeable performance boost if you sit there and pound the "VA Benefits and Health Care" button 🙂 

**If these changes are too much or are in bad timing please let me know and I'll reduce the scope to just what I need added.**

Later, I'd like to make sure we port over the updated vet-nav SCSS from vets-website, then start pulling from here instead. 

### Screenshots w/changes
image

See this ticket for more context, https://github.com/department-of-veterans-affairs/vets.gov-team/issues/17558.


packages/formation-react/src/components/MegaMenu/MegaMenu.unit.spec.jsx
```diff
@@ -182,7 +182,9 @@ describe('<MegaMenu>', () => {
   });
 
   it('should run toggleDropDown if clicked on', () => {
-    const spy = sinon.spy(megaMenu.instance(), 'toggleDropDown');
+    const spy = sinon.spy();
+
+    megaMenu.setProps({ toggleDropDown: spy })
```
ncksllvn: This was necessary now that the MegaMenu contains a child component, Navbar.

packages/formation-react/src/components/MegaMenu/Column.jsx
```diff
@@ -57,14 +61,15 @@ const Column = props => {
...
+        panelWhite,
+      )}`}
     >
-      <h3 id={`vetnav-${_.kebabCase(keyName)}-header`}>{data.title}</h3>
+      <h3 id={`vetnav-${ariaLabelledBy}-header`}>{data.title}</h3>
       <ul
-        id={`vetnav-${_.kebabCase(keyName)}-col`}
-        aria-labelledby={`vetnav-${_.kebabCase(keyName)}-header`}
+        id={`vetnav-${ariaLabelledBy}-col`}
+        aria-labelledby={`vetnav-${ariaLabelledBy}-header`}
```
ncksllvn: The ` aria-labelledby` has to be unique for the Axe check to pass. With the whole menu generating during startup, we had to use something other than the "columnOne" for that value, since each menu has columns.

PR #47, opened 2019-03-28
Update axe-core to latest version

jbalboni:
This updates us to the latest version of axe-core, as well as cleans up a couple of unit test warnings.

I also added the ability to ignore violations and added one such id to the errorable file input. Axe complains about us adding a button role to a label, which is fair, but this is currently working across browsers and screen readers, I believe. I'll create an issue and maybe @1Copenut can weigh in on if we should update this component.


1Copenut:
@jbalboni LMK what your issue number is for updating `ErrorableFileInput` and I'll keep an eye on it. The issue you're seeing can be remediated by adding a span inside the label to carry the `role="button"` attribute. I found a good article that outlines the process here: https://accessabilly.com/a-styled-accessible-file-upload/

LMK if that article won't work or you want to discuss more.

jbalboni:
@1Copenut Here's the ticket: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/17642

PR #48, opened 2019-03-28
[MegaMenu] Add new hubs without child pages

ncksllvn:
The addition of two new hubs with no child pages to VA.gov has introduced a new case for the MegaMenu - we should be able to add items into the sidebar of the "VA Benefits and Health Care" dropdown, but as plain links rather than section titles.

This PR also ports more vetnav CSS from vets-website, because it wasn't possible to make the changes otherwise without adding wild selectors.

## Desktop
image

## Mobile
image

## Testing
I tested this in about a million different ways but I would still say these changes are high risk. I could not confidently make any changes because of how wild our SCSS for this component has become.

### DIFF
--- a/packages/formation/sass/site/_m-vet-nav.scss
+++ b/packages/formation/sass/site/_m-vet-nav.scss
@@ -8,7 +8,7 @@ body.va-pos-fixed {
 }
 
 #vetnav,
-[role='menu'] [role='menu'] {
+[role="menu"] [role="menu"] {
   // The main menu takes up the whole screen under the Menu button for xsmall & small screens
   // Sub menus also take up the whole screen to cover the main menu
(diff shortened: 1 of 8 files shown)


packages/formation/sass/site/_m-vet-nav.scss
```diff
@@ -207,7 +207,7 @@ body.va-pos-fixed {
     }
   }
 
-  &[aria-expanded='true'] {
+  &[aria-expanded="true"] {
```
rianfowler: guess we need to add scss linting to this repo.

ncksllvn: Spoke to @jbalboni and moved the changes to this file over to the vet-website PR, https://github.com/department-of-veterans-affairs/vets-website/pull/9799/files.

PR #50, opened 2019-04-01
Upgrade USWDS to 1.6.10

ncksllvn:
Note - commit 84a45f902a9d30287ed08c1da8309efd68c3aa1e of this PR contains a change to publish an alpha version of VFST for easier peer review and will be reverted once peer review is done. This alpha version is consumed in https://github.com/department-of-veterans-affairs/vets-website/pull/9815. 

You can verify that you're seeing the correct USWDS version by inspecting a focusable element, applying the focus, state, then see that the focus rule is [no longer `*:focus`](https://user-images.githubusercontent.com/1915775/55360054-50963580-54a1-11e9-993c-b88c3a5be7db.png).

---
This PR contains some updates based on testing https://github.com/department-of-veterans-affairs/vets-website/pull/9815. This ticket is https://github.com/department-of-veterans-affairs/vets.gov-team/issues/17724.

Via `vets-website`'s visual regression testing as well as manual testing, the only changes I could identify were that the focus state applies only to focusable elements rather than globally. This meant we needed to update our own selectors for the focus state (to be the gold outline instead of the dotted outline) because the selectors were now more specific. Instead of copying over the USWDS selectors or applying a more specific one in ours, I just overrode the USWDS SCSS variable and mixin to achieve the same result.

#### Focus
image
image
image
image
image
image


packages/formation/sass/base/_b-variables.scss
```diff
@@ -87,6 +87,8 @@ $color-gold-light:           #f9c642;
 $color-visited:              $color-purple;
 $color-focus:                #3e94cf;
 
+$focus-outline:              2px solid $color-gold-light;
```
ncksllvn: By applying this variable override and the `@mixin focus` override, we don't need to have a whole file of overrides for the focus state.

packages/formation/sass/core.scss
```diff
@@ -27,6 +27,11 @@
 @import "~uswds/src/stylesheets/core/fonts";
 @import "~uswds/src/stylesheets/core/grid";
 @import "~uswds/src/stylesheets/core/utilities";
+
+// VA mixins. Here so that we can override USWDS mixins.
+// Specifically, the "focus" mixin.
+@import "base/b-mixins";
+
```
ncksllvn: This allows us to override the focus mixin. This seems like a pattern we should do more often, such as by overriding their [heading mixins](https://github.com/uswds/uswds/blob/c98f69c879d0fc9e2e2cc72ee9ee9743c1971d39/src/stylesheets/core/_utilities.scss#L7) with [ours ](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/master/packages/formation/sass/base/_b-mixins.scss#L27).

PR #51, opened 2019-04-02
fix linting issues

rianfowler:


packages/documentation/plugins/github-api-pages/github-api.js
```diff
@@ -6,32 +6,7 @@ const github = new GithubGraphQLApi({
   token: process.env.GITHUB_API_KEY,
 });
 
-const getPagesAndCreateNodes = async (pages, createNode) => {
-  return await Promise.all(pages.directoryPaths.map(async dir => {
-    if (path.extname(dir)) {
-      return await getPageAndCreateNode (
```
rianfowler: no return await is a newish thing to me but makes sense .

PR #53, opened 2019-04-03
Update Formation JS output (formation.js) to work standalone#17375

accbjt:
Currently, to use accordions and interactive components in formation from the USWDS, a user of formation must install USWDS as a dependency in order to access the USWDS javascript.

Now, we want to enable users of formation to use accordions and interactive components without USWDS as a dependency. This work will also establish the basis for which other interactive formation components will be implemented in vanilla javascript.


accbjt:
Let me know if I am moving in the right direction. What I did to test it out is I created a new project and within my project I added the accordian html and then did an `import '@department-of-veterans-affairs/formation/dist/formation'`. This should also work if you don't use node_modules and add it as a script tag. I used yalc for testing the module. Not sure if this is the best approach or not. I'm welcome to any suggestions.

rianfowler:
Alternatively we could try to use local overrides on the uswds.js file: https://developers.google.com/web/updates/2018/01/devtools#overrides

that would probably be simpler - would be happy to help you with this.

accbjt:
@rianfowler I'll try it on vets-website and see if It works. Thanks

packages/formation/js/accordion.js
```diff
@@ -0,0 +1,59 @@
+const toBoolean = value => value === 'true';
```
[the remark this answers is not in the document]
accbjt: Looks like I am missing some viewport stuff but I think most of it is there. I also made sure to look at their tests also to see if I was missing any. I also added the tests to formation so we don't lose any functionality.

packages/formation/package.json
```diff
@@ -19,15 +19,18 @@
...
+    "test": "jest"
   },
   "bugs": {
     "url": "https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/issues"
   },
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.6.3",
     "foundation-sites": "5",
-    "moment": "^2.23.0"
+    "jest": "^24.7.1",
```
jbalboni: I think all three of these should be in devDependencies.

jbalboni:
@accbjt I think we should set up the way the accordion component is initialized differently.

As it stands now, if you import the accordion file in another site it'll automatically add the handlers when the DOM is ready. This is going to make it hard for us to migrate away from using the uswds script tag on teamsite pages, because I believe we'll need to conditionally attach the handler depending on if uswds is already a script on the page. That way we can update our code, then request changes to the injected scripts block, without the site breaking.

This is what I would suggest: make `loadAccordionHandler` an exported function. And then create another js file that will serve as the bundle entry point in webpack that imports `loadAccordionHandler`, checks if the DOM is ready and calls `loadAccordionHandler` when it is.

That way we can import that root entry point in vets-website for our normal pages, but conditionally execute `loadAccordionHandler` (and any future components we need) in the teamsite bundle.

packages/formation/js/onDocumentReady.js
```diff
@@ -0,0 +1,10 @@
+import loadAccordionHandler from './accordion';
```
accbjt: @jbalboni Let me know if this is what you were suggesting to do with initializing the accordion.

jbalboni:
@accbjt Can you remove the yarn.lock file that was added to the formation package? I don't think we need that, Lerna manages those dependencies without it.

packages/formation/.babelrc
```diff
@@ -0,0 +1,12 @@
+{
+  "presets": [
+    [
+      "babel-preset-env",
+      {
+        "targets": {
+          "node": "current"
```
jbalboni: You should copy the settings from vets-website's babelrc, I think.

webpack.config.prod.js
```diff
@@ -1,12 +1,13 @@
 const path = require('path');
+const glob = require('glob');
```
jbalboni: I think you can remove this

packages/formation/.babelrc
accbjt: Sure

jbalboni: Probably not the React ones, but I think this would also be used for the formation.js output, so it would need to have the same environment targets and ES2015+ proposal plugins.

accbjt: Okay, sound good.

packages/formation/.babelrc
```diff
@@ -0,0 +1,57 @@
...
+        /* This is the full list of browsers we support, but we are not using it because we want babel-polyfill
+         * to include only polyfills for more modern browsers.
+         * Older browser polyfills are in preESModulesPolyfills.js
+         */
+        /* "browsers": ["Chrome 60", "Firefox 57", "iOS 9", "Edge 14", "ChromeAndroid 64", "Safari 10", "ie 11"] */
+        /*
+         * This is the list of browers we support that all also have support for the nomodule script
+         * attribute, meaning we can make them ignore a file that includes polyfills for older browsers
+         */
+        "browsers": ["Chrome 61", "Firefox 60", "iOS 11", "Edge 16", "ChromeAndroid 67", "Safari 11"]
```
jbalboni: Lets actually remove this browser list and use the commented out one above. I don't think it'll make an actual difference, but we don't have the same polyfill set up here, so it's safer to use the other list.

jbalboni:
Closing in favor of reset PR: https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/73

PR #76, opened 2019-04-17
Remove topology doc

jbalboni:
Remove topology doc for now.


gunsch:
does this need to include removing the image assets? where are those pulled in from?

jbalboni:
@gunsch Yes, good catch. Those are removed now.

PR #81, opened 2019-04-23
Rearrange platform and architecture docs

jbalboni:
I reorganized some content here between the architecture and platform sections and ended up removing architecture entirely, in favor of nesting it under platform. I made a new section called Visual design, which contains the formation components, and probably some future style information.

Screen Shot 2019-04-23 at 3 16 43 PM


jbalboni:
@rianfowler Definitely. Architecture is a pretty vague term, maybe we could get to having specific groups of content and do away with it entirely.

PR #82, opened 2019-04-24
Accordion errors in Formation 6.2.0: Bug/18077

accbjt:
The script does not work on bordered accordions as they are coded in USWDS 1.x: 

There appears to be issues with traversing the DOM. Notice when I click the top set of accordions, the bottom set collapses.

### DIFF
--- a/packages/formation/__tests__/js/accordion/accordion.test.js
+++ b/packages/formation/__tests__/js/accordion/accordion.test.js
@@ -204,7 +204,7 @@ describe('accordion', () => {
     );
   });
 
-  it('.usa-accordion click should not effect .usa-accordion-bordered when clicked', () => {
+  it('.usa-accordion click should not affect .usa-accordion-bordered when clicked', () => {
     require('../../../js');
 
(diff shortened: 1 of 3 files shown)


jaredcunha-usds:
The accordions now seem to work in the design system site. 

One thing I found a little jarring was the `scrollIntoView` behavior. 
Screen Recording 2019-04-24 at 10 29 AM

jaredcunha-usds:
About my comment above, since it's unrelated to the bug, maybe that can be a 6.2.2 thing?

Take a look at the example on USWDS. It seems to scroll the page if the button is cut off… not the content. This is from https://v2.designsystem.digital.gov/components/accordion/
Screen Recording 2019-04-24 at 11 43 AM

accbjt:
@jaredcunha-usds I fixed the scroll issue. Can you try it again and let me know if it's good.

jaredcunha-usds:
Behaving as expected.
Screen Recording 2019-04-25 at 11 30 AM

accbjt:
@jaredcunha-usds Thanks, I will merge this in and publish it.

PR #84, opened 2019-04-24
elastic lunr

rianfowler:


packages/documentation/src/components/Header.jsx
```diff
@@ -1,7 +1,63 @@
+/* eslint-disable react/jsx-key */
 import React from 'react';
-import { Link } from 'gatsby';
+import { Link, graphql } from 'gatsby';
 import sidebarData from '../sidebar';
 import MobileNav from './MobileNav';
+import { Index } from 'elasticlunr';
+// Graphql query used to retrieve the serialized search index.
+export const query = graphql`
```
rianfowler: the documentation doesn't say where this comes from or what to do with it. the example uses [this](https://github.com/andrew-codes/gatsby-plugin-elasticlunr-search-demo/blob/6f662e66ce0ad48f1a0bc4bf6579a618958deb29/src/templates/default.js#L1)

packages/documentation/src/components/Header.jsx
```diff
@@ -1,7 +1,63 @@
...
+        </ul>
+      </div>
+    );
+  }
+
+  getOrCreateIndex = () => {
+    return this.index
+      ? this.index
+      : // Create an elastic lunr index and hydrate with graphql query results
+        Index.load(this.props.data.siteSearchIndex.index);
```
rianfowler: This blows up.
Screen Shot 2019-04-24 at 13 09 31

The documentation doesn't say where this prop is supposed to come from. I traced it through the [example](https://github.com/andrew-codes/gatsby-plugin-elasticlunr-search-demo/blob/6f662e66ce0ad48f1a0bc4bf6579a618958deb29/src/templates/default.js#L8) and it's not clear where any of this data comes from.

packages/documentation/src/pages/getting-started/workflow/deploy.mdx
```diff
@@ -1,3 +1,8 @@
+---
+title: deploy
+keywords: cat, dog, spices
```
rianfowler: documentation suggests that front matter is indexed and searchable

packages/documentation/package.json
```diff
@@ -39,6 +40,9 @@
     "gatsby"
   ],
   "license": "MIT",
+  "resolutions": {
+    "graphql": "14.1.1"
```
rianfowler: gatsby blows up after installing this plugin with an error about multiple versions of graphql. i added this to force all of the dependencies to use this version.

packages/documentation/gatsby-config.js
```diff
@@ -66,5 +66,23 @@ module.exports = {
     `gatsby-transformer-remark`,
     `github-api-pages`,
     `mdx-pages`,
+    {
+      resolve: `@andrew-codes/gatsby-plugin-elasticlunr-search`,
```
jbalboni: You might want to try https://github.com/gatsby-contrib/gatsby-plugin-elasticlunr-search. It looks like it's a fork that will work with Gatsby 2, which we're on.

packages/documentation/gatsby-config.js
```diff
@@ -10,6 +10,26 @@ module.exports = {
...
+      resolve: `@gatsby-contrib/gatsby-plugin-elasticlunr-search`,
+      options: {
+        // Fields to index
+        fields: [`title`, `tags`],
+        // How to resolve each field`s value for a supported node type
+        resolvers: {
+          // For any node of type MarkdownRemark, list how to resolve the fields` values
+          MarkdownRemark: {
+            title: node => {
+              console.log(node);
```
rianfowler: I think everything is wired up correctly but the search is returning no results. I don't really understand the gatsby pipeline but AFAIK none of the nodes that this gets are the mdx content. It's mostly just comments from jsx files. 

Screen Shot 2019-04-25 at 14 37 22

I tried to delete my cache and find content from the mdx files I'd added front matter to but no luck. 

My understanding is that this needs access to the content nodes so it can build an index from the front matter fields.

packages/documentation/gatsby-config.js
```diff
@@ -86,5 +66,51 @@ module.exports = {
...
+            tags: node => {
+              if (node.context && node.context.frontmatter && node.context.frontmatter.tags) {
+                return node.context.frontmatter.tags;
+              }
+
+              return '';
+            },
+            path: node => node.path,
+          },
+          Mdx: {
```
rianfowler: Need to target specific node types. I don't know if `SitePage` and `Mdx` cover everything we want to index. Here's a list of all of the node types:

```'SitePage',
  'SitePlugin',
  'Site',
  'Directory',
  'File',
  'ImageSharp',
  'ComponentProp',
  'ComponentMetadata',
  'ComponentDescription',
  'Mdx',
  'MarkdownRemark',
  'SiteSearchIndex'
```

packages/documentation/gatsby-config.js
```diff
@@ -86,5 +66,51 @@ module.exports = {
...
+              return '';
+            },
+            tags: node => {
+              if (node && node.frontmatter && node.frontmatter.tags) {
+                return node.frontmatter.tags;
+              }
+
+              return '';
+            },
+            path: node => `/visual-design/components${node.fields.slug}`,
```
rianfowler: not a fan of this- I didn't see anything on the node I could pull off to get the path. maybe there's some graphQLing I could do to get the path?

packages/documentation/src/components/Header.jsx
```diff
@@ -37,17 +35,30 @@ class Search extends React.Component {
...
+        />
+        <ul
+          id="results-container"
+          className="site-search-results"
+          role="listbox"
+        >
           {this.state.results.map(page => (
             <li key={page.id}>
               <Link to={`/${page.path}`}>{page.title}</Link>
-              {`: ${page.tags.join(`,`)}`}
```
rianfowler: Left the tags off for now. There's a lot we can do with this area to make the search nice. Would like some ideas- maybe we try to format it to look similar to the docusearch? @jbalboni @accbjt if you have other searches you like share them with me. 

I am going to add a lot of front matter to get the indexing working right. like I don't want people to have to type "Errorable" before seeing any of the components so I'll add "checkbox" to "ErrorableCheckbox"'s tags. I might break that off into a separate ticket so we can have a PR that's just focused on the indexing. 

Screen Shot 2019-04-25 at 17 09 15

packages/documentation/src/components/Header.jsx
```diff
@@ -93,21 +104,18 @@ export default class Header extends React.Component {
...
-                    placeholder="Search..."
-                    aria-label="Search"
-                  />
-                  <ul
-                    id="results-container"
-                    className="site-search-results"
-                    role="listbox"
-                  />
-                </div>
+                <StaticQuery
```
rianfowler: This component composition is a little messy- I might move the search out of this file and just export a functional component with this composition.

packages/documentation/src/pages/getting-started/common-tasks/test.mdx
```diff
@@ -1,3 +1,8 @@
+---
+title: Run tests
+tags: unit test, lint, end to end test, accessibility, 508
```
rianfowler: we can add whatever we want to the front matter for indexing. I think it would be nice to index the section titles too but I didn't see them easily available in the nodes.

packages/documentation/gatsby-config.js
```diff
@@ -86,5 +66,51 @@ module.exports = {
...
+      resolve: `@gatsby-contrib/gatsby-plugin-elasticlunr-search`,
+      options: {
+        // Fields to index
+        fields: [`title`, `tags`],
+        // How to resolve each field`s value for a supported node type
+        resolvers: {
+          // For any node of type MarkdownRemark, list how to resolve the fields` values
+          SitePage: {
+            title: node => {
+              if (node.context && node.context.frontmatter && node.context.frontmatter.title) {
```
rianfowler: lodash get or optional chaining would make this cleaner- not sure how to fix this upstream when the nodes are created

packages/documentation/src/components/Header.jsx
rianfowler: I'll limit this list size so it doesn't scroll

packages/documentation/gatsby-config.js
jbalboni: There's a query in SidebarItems for getting the list of component page paths for the sidebar. Maybe that would help?

packages/documentation/plugins/mdx-pages/gatsby-node.js
```diff
@@ -66,6 +67,9 @@ exports.createPages = async ({ graphql, actions }) => {
         context: {
           id: node.id,
           source: 'component',
+          frontmatter: {
+            ...node.frontmatter,
+          },
           name: node.frontmatter.name,
```
rianfowler: Turns out I don't need to query the Mdx nodes- I needed to expand the front matter query and then add it to the context. I'll probably need to expand this to include tags but this fixes the hardcoded path issue.

packages/documentation/src/components/Header.jsx
```diff
@@ -1,7 +1,140 @@
...
+          .splice(0, maxDisplayedSearchResults)
+          .map(({ ref }) => {
+            const result = this.index.documentStore.getDoc(ref);
+
+            return {
+              ...result,
+              filteredTags: result.tags
+                .split(',')
+                .filter(tag =>
+                  query.split(' ').find(queryPart => tag.includes(queryPart)),
```
rianfowler: tags are matched across multiple words 

Screen Shot 2019-04-30 at 17 37 55
Screen Shot 2019-04-30 at 17 37 49

packages/documentation/src/components/Header.jsx
```diff
@@ -1,7 +1,140 @@
...
+  getOrCreateIndex() {
+    return this.index
+      ? this.index
+      : // Create an elastic lunr index and hydrate with graphql query results
+        Index.load(this.props.searchIndex);
+  }
+
+  search(evt) {
+    const query = evt.target.value;
+    if (query.length > 2) {
```
rianfowler: doesn't search until there are 3 or more letters. may want to have a search results page eventually

packages/documentation/src/components/Header.jsx
```diff
@@ -1,7 +1,140 @@
...
+      this.index = this.getOrCreateIndex();
+      this.setState({
+        query,
+        // Query the index with search string to get an [] of IDs
+        results: this.index
+          .search(query, { expand: true })
+          // Map over each ID and return the full document
+          .splice(0, maxDisplayedSearchResults)
```
rianfowler: only showing 5 results currently

packages/documentation/src/components/Header.jsx
```diff
@@ -1,7 +1,140 @@
...
+    return <>{string}</>;
+  }
+
+  const queryStartIndex = lowerCasedString.indexOf(matchedQuery);
+
+  return (
+    <>
+      {string.slice(0, queryStartIndex)}
+      <strong>
+        {string.slice(queryStartIndex, queryStartIndex + matchedQuery.length)}
```
rianfowler: multiple words are treated as multiple queries and the component bolds the first matching query. 

Screen Shot 2019-04-30 at 18 11 06
Screen Shot 2019-04-30 at 18 11 14

packages/documentation/plugins/mdx-pages/gatsby-node.js
```diff
@@ -66,6 +68,9 @@ exports.createPages = async ({ graphql, actions }) => {
         context: {
           id: node.id,
           source: 'component',
+          frontmatter: {
+            ...node.frontmatter,
```
rianfowler: should only need to add new front matter to the query on the MDX components with this.

packages/documentation/gatsby-config.js
```diff
@@ -66,5 +66,35 @@ module.exports = {
...
+        fields: [`title`, `tags`],
+        // How to resolve each field`s value for a supported node type
+        resolvers: {
+          // For any node of type MarkdownRemark, list how to resolve the fields` values
+          SitePage: {
+            title: (node, getNode) => {
+              if (node.context && node.context.frontmatter && node.context.frontmatter.title) {
+                return node.context.frontmatter.title;
+              } else {
+                console.info('Page title missing from front matter.', node);
```
rianfowler: I think we should have some warning for pages that are missing frontmatter- they won't be indexed without this.

rianfowler:
You may need to delete your gatsby cache for the new front matter to get indexed.

packages/documentation/src/components/Header.jsx
```diff
@@ -1,7 +1,140 @@
...
+      </strong>
+      {string.slice(queryStartIndex + matchedQuery.length)}
+    </>
+  );
+}
+
+class Search extends React.Component {
```
jbalboni: Could we move this component to a separate file?

PR #86, opened 2019-04-25
Add optional `backgroundOnly` prop to AlertBox

erikphansen:
Seemed like it should be possible to turn an AlertBox into its 'background only' variant without resorting to passing in a class name like I've had to do [here](https://github.com/department-of-veterans-affairs/vets-website/blob/d1afbcd92d91d4892469277ca1b96163f2781946/src/applications/personalization/dashboard/components/ManageYourVAHealthCare.jsx#L73) and [here](https://github.com/department-of-veterans-affairs/vets-website/blob/3e7daca36ae04b7263eb12f0d889c75f305391ed/src/applications/hca/containers/IntroductionPageGated.jsx#L96)

### DIFF
--- a/packages/formation-react/src/components/AlertBox/AlertBox.jsx
+++ b/packages/formation-react/src/components/AlertBox/AlertBox.jsx
@@ -38,6 +38,7 @@ class AlertBox extends React.Component {
     const alertClass = classNames(
       'usa-alert',
       `usa-alert-${this.props.status}`,
+      { 'background-color-only': this.props.backgroundOnly },
       this.props.className,
     );
 
@@ -115,10 +116,17 @@ AlertBox.propTypes = {
    * Optional class name to add to the alert box.
    */
   className: PropTypes.string,
+
+  /**
+   * If true, renders an AlertBox with only a background color, without an
+   * accented left edge or an icon
+   */
+  backgroundOnly: PropTypes.bool,
 };
 
 AlertBox.defaultProps = {
   isVisible: true,
+  backgroundOnly: false,
 };
 
 export default AlertBox;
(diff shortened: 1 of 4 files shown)


packages/formation-react/src/components/AlertBox/AlertBox.jsx
```diff
@@ -38,6 +38,7 @@ class AlertBox extends React.Component {
     const alertClass = classNames(
       'usa-alert',
       `usa-alert-${this.props.status}`,
+      { 'background-color-only': this.props.backgroundOnly },
```
ncksllvn: I haven't seen this `background-color-only` class before - should it have the VADS prefix?

packages/formation-react/src/components/AlertBox/AlertBox.unit.spec.jsx
```diff
@@ -21,13 +21,34 @@ describe('<AlertBox />', () => {
...
     expect(wrapper.find('.usa-alert').hasClass('usa-alert-info')).to.equal(
       true,
     );
+    expect(
+      wrapper.find('.usa-alert').hasClass('background-color-only'),
+    ).to.equal(false);
+    wrapper.unmount();
+  });
+
+  it('should have have `background-color-only` class added when `backgroundOnly` is `true`', () => {
```
jbalboni: Can we add another test to run aXe with the background color prop set? There's a group of axeCheck tests at the end of this file.

packages/formation-react/src/components/AlertBox/AlertBox.jsx
jbalboni: I think it predates the `vads-` convention.

packages/formation-react/src/components/AlertBox/AlertBox.unit.spec.jsx
erikphansen: Done

PR #89, opened 2019-04-26
Set Babel to transpile modules

jbalboni:
When we updated the Babel config I accidentally carried over the `modules: false` option from the vets-website config, which is not what we want in this repo. That caused `imports` to stop being converted to `require`s.

To test this locally, you'll need to run the build command once with `BABEL_DISABLE_CACHE=1` set as an env variable.

### DIFF
--- a/packages/formation-react/scripts/build.js
+++ b/packages/formation-react/scripts/build.js
@@ -19,9 +19,9 @@ function flattenRequires(bufferString) {
     recast.print(
       recast.visit(recast.parse(bufferString), {
         visitCallExpression: function(path) {
+          this.traverse(path);
           var expr = path.node;
           if (expr.callee.name == 'require') {
-            this.traverse(path);
             if (
               expr.arguments.length &&
               expr.arguments[0].value.charAt(0) == '.'
--- a/yarn.lock
+++ b/yarn.lock
@@ -12426,7 +12426,7 @@ uuid@^3.0.1, uuid@^3.3.2:
     eslint-plugin-react "^7.12.4"
     eslint-plugin-react-hooks "^1.6.0"
     eslint-plugin-scanjs-rules "^0.1.4"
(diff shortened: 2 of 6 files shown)


packages/formation-react/scripts/build.js
```diff
@@ -19,9 +19,9 @@ function flattenRequires(bufferString) {
     recast.print(
       recast.visit(recast.parse(bufferString), {
         visitCallExpression: function(path) {
+          this.traverse(path);
```
jbalboni: The way this apparently works is to visit all top level call expressions, then `traverse` will traverse deeper into the AST and find more more call expressions. This is fine when all the require statements are at the first level of the AST, but after the recent Babel update, some of them are nested in an interop function. This change makes it so that all call expressions are traversed.

yarn.lock
```diff
@@ -12426,7 +12426,7 @@ uuid@^3.0.1, uuid@^3.3.2:
     eslint-plugin-react "^7.12.4"
     eslint-plugin-react-hooks "^1.6.0"
     eslint-plugin-scanjs-rules "^0.1.4"
-    eslint-plugin-va-enzyme "file:../../Library/Caches/Yarn/v4/npm-vagov-eslint-1.0.0-4e4c7dca-961b-4a86-a8d1-85502669dab7-1555430367307/node_modules/vagov-eslint/custom-eslint/va-enzyme"
+    eslint-plugin-va-enzyme "file:../../Library/Caches/Yarn/v4/npm-vagov-eslint-1.0.0-502d90db-d4ff-4ec6-9676-2b8486c70c63-1556302093236/node_modules/vagov-eslint/custom-eslint/va-enzyme"
```
erikphansen: weird, but ok

jbalboni: Yeah, this is an artifact of how we're copying around linting configs. I'd like to see if we can get rid of it eventually, but @rianfowler knows more about it.

rianfowler: I think that's the idiomatic way unless we want to publish it to npm

PR #90, opened 2019-04-29
Accordion bug in Formation: Bug/18238

accbjt:
Accordion not working correctly on mobile. Nested div's do not get recognized.

this works:
```
<div class="usa-accordion">
   <button class="usa-accordion-button">
   <div class="usa-accordion-content">
```

this doesn’t, but used to:
```
<div class="usa-accordion">
   <div>
        <button class="usa-accordion-button">
    </div>
   <div class="usa-accordion-content">
```


packages/documentation/src/pages/platform/unit-testing.mdx
```diff
@@ -0,0 +1,108 @@
+## How do unit tests work?
```
rianfowler: looks like some unit test content made it into this PR. 

you can clear out these changes with something like `git checkout origin master packages/documentation/src/pages/platform/unit-testing.mdx`

just make sure these changes are on your unit test branch.

packages/formation/js/accordion.js
```diff
@@ -26,9 +26,13 @@ const loadAccordionHandler = () => {
 
   usaAccordionContentElements.forEach(el => {
     if (el.getAttribute('aria-hidden') === null) {
+      const buttonElement = el.parentElement.querySelector(
```
rianfowler: could use `querySelectorAll` for some of this logic to make it more declarative. The query is kind of long but this logic is also really complicated.

```
// creates exactly the array you're looking for without using spread operator 
const usaAccordion = document.querySelectorAll('.usa-accordion, .usa-accordion-bordered')

// inverts the logic a little- looks for the button first and then queries the siblings
// the matching is done with the query e.g. where button's aria-expanded is true and .usa-accordion-content is not false 
document
.querySelectorAll('button[aria-controls][aria-expanded='true'] ~ .usa-accordion-content:not[aria-hidden='false')
.forEach(el => el.setAttribute(
	        'aria-hidden',
	        false,
	      );

document
.querySelectorAll('button[aria-controls][aria-expanded='false'] ~ .usa-accordion-content:not[aria-hidden='true')
.forEach(el => el.setAttribute(
	        'aria-hidden',
	        true,
	      );
```

I think this is mostly a matter of taste- I think query selectors for this kind of thing are easier to understand than a few lines of imperative code. Maybe @jbalboni has an opinion. 

Not totally relevant but `querySelectorAll` is supposed to be highly optimized.

accbjt: I didn't know you can even do that. Thanks

jaredcunha-usds:
Hmmm - As is, this is still not working on an older version of the mobile nav. Doesn't seem to impact anything else.

accbjt:
@jaredcunha-usds Can you share with me the html for the mobile nav your testing it on.

packages/formation/js/accordion.js
```diff
@@ -16,60 +16,63 @@ const isElementInViewport = (
...
 
 const loadAccordionHandler = () => {
-  const usaAccordion = [
-    ...document.getElementsByClassName('usa-accordion'),
-    ...document.getElementsByClassName('usa-accordion-bordered'),
-  ];
-  const usaAccordionContentElements = [
-    ...document.getElementsByClassName('usa-accordion-content'),
-  ];
+  document
```
accbjt: @rianfowler I used the query select but couldn't use your selector due to the situation when a button is nested in an element as how it is in mobile view on vets. The sibling combinator wouldn't work.  I couldn't find a selector that would work with nested children, so this is what I came up with. Any suggestions are welcomed.

packages/formation/js/accordion.js
```diff
@@ -16,60 +16,63 @@ const isElementInViewport = (
...
+    });
 
-  usaAccordion.forEach(element => {
-    element.addEventListener('click', e => {
-      const target = e.target;
-      const other = Array.from(
-        element.getElementsByClassName('usa-accordion-button'),
-      ).filter(item => item !== target);
+  document
+    .querySelectorAll('.usa-accordion, .usa-accordion-bordered')
```
accbjt: This is a lot better then using the spread operator. I didn't know querySelectorAll returned an element that had a forEach method in it. Thank @rianfowler

accbjt:
@jaredcunha-usds This should work now. I tested it on your repo.

jaredcunha-usds:
Still not opening: 
```
TypeError: e.parentElement.querySelector(...) is null[Learn More] formation.js:1:1638
    o http://localhost:4000/vendor/javascripts/formation.js:1
    forEach self-hosted:262
    o http://localhost:4000/vendor/javascripts/formation.js:1
```

accbjt:
@jaredcunha-usds yeah I must have added back the old code by accident when refactoring . Try it now.

jaredcunha-usds:
Ah okay that seems to work now!

jaredcunha-usds:
I think it' ll just need to be bumped to `6.2.4`

PR #94, opened 2019-05-02
16114 dashboard homeless alert screenreader changes

tlei123:
\[Companion to vet-website [PR10024](/department-of-veterans-affairs/vets-website/pull/10024) (Pass more descriptive aria-label string to AlertBox)\]
[16114](/department-of-veterans-affairs/vets.gov-team/issues/16114) - Change AlertBox aria attributes:

- Refactor close button aria-label to optionally accept a more descriptive string-value.
- Remove the aria-live attribute from outer element.  React removes the entire alert element when User clicks the close button, so this attribute (which announces inner-content updates) is moot.

Also bumped formation-react version to 4.4.3.

### DIFF
--- a/packages/formation-react/package.json
+++ b/packages/formation-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "4.4.3",
+  "version": "4.5.0",
   "description": "React implementation of patterns in Formation",
   "keywords": [
     "react",
--- a/packages/formation-react/package.json
+++ b/packages/formation-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "4.4.2",
+  "version": "4.4.3",
   "description": "React implementation of patterns in Formation",
   "keywords": [
     "react",
--- a/packages/formation-react/src/components/AlertBox/AlertBox.jsx
+++ b/packages/formation-react/src/components/AlertBox/AlertBox.jsx
@@ -45,7 +45,7 @@ class AlertBox extends React.Component {
     const closeButton = this.props.onCloseAlert && (
       <button
         className="va-alert-close"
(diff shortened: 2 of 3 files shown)


packages/formation-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "4.4.2",
+  "version": "4.4.3",
```
erikphansen: let's make this `4.5.0` since it's adding a feature and isn't just a bug fix

packages/formation-react/src/components/AlertBox/AlertBox.jsx
```diff
@@ -45,7 +45,7 @@ class AlertBox extends React.Component {
     const closeButton = this.props.onCloseAlert && (
       <button
         className="va-alert-close"
-        aria-label="Close notification"
+        aria-label={this.props.closeBtnAriaLabel || 'Close notification'}
```
erikphansen: I would be curious what others think about setting a default value for the `closeBtnAriaLabel` prop instead of doing the conditional here.

rianfowler: That would be more idiomatic.

jbalboni: Would also show up in the component docs if we do it that way.

packages/formation-react/package.json
tlei123: Bumped now to 4.5.0.

packages/formation-react/src/components/AlertBox/AlertBox.jsx
tlei123: Added default prop & aXe test.

tlei123:
@erikphansen , in cash this doesn't get Approved 'til tomorrow (Friday -- I‘ll be on PTO)...

If you have a chance, please also merge & publish.

Thanks much!

PR #95, opened 2019-05-02
Fix Accordion Click event collision: Bug/18359

accbjt:
When adding the formation.js there is a problem with React components click events and the events that are created with this code. We need to check whether an element already has a click event assigned to it and not run the accordion code if it does. This was discovered when implementing it on vets-website.

### DIFF
--- a/packages/formation/js/accordion.js
+++ b/packages/formation/js/accordion.js
@@ -15,7 +15,7 @@ const isElementInViewport = (
   );
 };
 
-const loadAccordionHandler = () => {
+const addAriaHiddenAttr = () => {
   document
     .querySelectorAll('.usa-accordion-content:not([aria-hidden])')
     .forEach(el => {
@@ -28,7 +28,14 @@ const loadAccordionHandler = () => {
 
       el.setAttribute('aria-hidden', hiddenValue);
     });
+};
 
+const getOtherButtons = (element, target) =>
+  Array.from(element.getElementsByClassName('usa-accordion-button')).filter(
+    item => item !== target,
+  );
+
+const addAccordionClickHandler = () => {
   document
     .querySelectorAll('.usa-accordion, .usa-accordion-bordered')
     .forEach(element => {
@@ -38,27 +45,25 @@ const loadAccordionHandler = () => {
         // Checks whether the button has a click event already assigned to it.
         // Specifically React Components.
         if (!target.onclick) {
-          const other = Array.from(
-            element.getElementsByClassName('usa-accordion-button'),
-          ).filter(item => item !== target);
-
           const multiSelectable = toBoolean(
             element.getAttribute('aria-multiselectable'),
           );
 
-          if (target.getAttribute('aria-controls')) {
-            if (!multiSelectable) {
-              other.forEach(el => {
-                const contentEl = el.getAttribute('aria-controls');
+          const hasAriaControlsAttr = target.getAttribute('aria-controls');
 
-                el.setAttribute('aria-expanded', 'false');
+          if (hasAriaControlsAttr && !multiSelectable) {
+            getOtherButtons(element, target).forEach(el => {
+              const contentEl = el.getAttribute('aria-controls');
 
-                document
-                  .getElementById(contentEl)
-                  .setAttribute('aria-hidden', 'true');
-              });
-            }
+              el.setAttribute('aria-expanded', 'false');
 
+              document
+                .getElementById(contentEl)
+                .setAttribute('aria-hidden', 'true');
+            });
+          }
+
+          if (hasAriaControlsAttr) {
             const dropDownElement = document.getElementById(
               target.getAttribute('aria-controls'),
             );
@@ -76,6 +81,11 @@ const loadAccordionHandler = () => {
         }
       });
     });
+};
+
+const loadAccordionHandler = () => {
+  addAriaHiddenAttr();
+  addAccordionClickHandler();
 };
 
 export default loadAccordionHandler;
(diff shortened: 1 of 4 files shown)


packages/formation/js/index.js
```diff
@@ -1,10 +1,13 @@
...
+import domready from 'domready';
 
 if (
   document.readyState === 'complete' &&
   (document.readyState !== 'loading' && !document.documentElement.doScroll)
 ) {
   loadAccordionHandler();
 } else {
-  document.addEventListener('DOMContentLoaded', loadAccordionHandler);
+  domready(() => {
```
accbjt: added this because I had issues when using this on vets-website. I saw this is what  uswds uses, so I tried it out. So far it works.

packages/formation/js/accordion.js
```diff
@@ -34,39 +34,44 @@ const loadAccordionHandler = () => {
...
-          element.getElementsByClassName('usa-accordion-button'),
-        ).filter(item => item !== target);
 
-        const multiSelectable = toBoolean(
-          element.getAttribute('aria-multiselectable'),
-        );
+        // Checks whether the button has a click event already assigned to it.
+        // Specifically React Components.
+        if (!target.onclick) {
+          const other = Array.from(
```
rianfowler: What is `other`?

packages/formation/js/accordion.js
```diff
@@ -34,39 +34,44 @@ const loadAccordionHandler = () => {
...
-        if (target.getAttribute('aria-controls')) {
-          if (!multiSelectable) {
-            other.forEach(el => {
-              const contentEl = el.getAttribute('aria-controls');
+          const multiSelectable = toBoolean(
+            element.getAttribute('aria-multiselectable'),
+          );
 
-              el.setAttribute('aria-expanded', 'false');
+          if (target.getAttribute('aria-controls')) {
```
rianfowler: Let's consolidate this conditional. Maybe something like
```
const isMultiSelectable = toBoolean(
  element.getAttribute('aria-multiselectable'),
);

const hasAriaControlsAttribute = toBoolean(target.getAttribute('aria-controls'));

if (hasAriaControlsAttribute && ! isMultiSelectable) {

```

packages/formation/js/accordion.js
```diff
@@ -34,39 +34,44 @@ const loadAccordionHandler = () => {
...
-          const targetExpandedAttr = toBoolean(
-            target.getAttribute('aria-expanded'),
-          );
+                document
+                  .getElementById(contentEl)
+                  .setAttribute('aria-hidden', 'true');
+              });
+            }
+
+            const dropDownElement = document.getElementById(
```
rianfowler: are we sure that there will only ever be one `aria-controls` on the page?

packages/formation/js/accordion.js
```diff
@@ -34,39 +34,44 @@ const loadAccordionHandler = () => {
...
+              target.getAttribute('aria-controls'),
+            );
+            const targetExpandedAttr = toBoolean(
```
rianfowler: Since this typecast to boolean (and isn't the attribute itself) it might be better named as `hasTargetExpandedAttribute`

packages/formation/js/index.js
rianfowler: can probably just call `domready(loadAccordionHandler)`

packages/formation/js/accordion.js
```diff
@@ -34,39 +34,44 @@ const loadAccordionHandler = () => {
...
-            other.forEach(el => {
-              const contentEl = el.getAttribute('aria-controls');
+          const multiSelectable = toBoolean(
+            element.getAttribute('aria-multiselectable'),
+          );
 
-              el.setAttribute('aria-expanded', 'false');
+          if (target.getAttribute('aria-controls')) {
+            if (!multiSelectable) {
+              other.forEach(el => {
```
rianfowler: what is `other`?

packages/formation/js/accordion.js
accbjt: @rianfowler other is the other usa-accordion-buttons. Would otherButtons be better?

packages/formation/js/accordion.js
accbjt: @rianfowler I'm not sure but under the target element there should only be 1. I guess the only situation where there might be another is a nested accordion. Do we want to try to handle this use case.

packages/formation/js/index.js
accbjt: Yeah your right

packages/formation/js/accordion.js
rianfowler: are they nested or siblings? is it one or many?

packages/formation/js/accordion.js
rianfowler: I see now- we're getting the id for that control which is unique and using that to query the element

packages/formation/js/accordion.js
accbjt: There are many. The current target is the <ul> accordion element and it will get the usa-accordion-button child elements. The child button element are inside of the <li> element but can be nested another level down. So they are not direct siblings.

packages/formation/js/accordion.js
accbjt: Actually the problem is that this gives us back a `"true"` value as a string and I need to convert it into a boolean, so I can do the opposite using the `!targetExpandedAttr`. So it actually gives us back true and false but because it is html it gives it back as a string.

accbjt:
@rianfowler I did some refactoring to make it more readable. Let me know what you think.

PR #96, opened 2019-05-02
Add docs for metalsmith flags

jbalboni:
Documentation that goes along with https://github.com/department-of-veterans-affairs/vets-website/pull/10016

### DIFF
--- a/packages/documentation/src/pages/platform/tools/feature-flags.mdx
+++ b/packages/documentation/src/pages/platform/tools/feature-flags.mdx
@@ -1,6 +1,4 @@
-# Front End Feature Flags
-
-## Overview
+# Front end feature flags
 
 Agile and user centric workflows often use testing of partially complete features to refine or research experiences. This document provides recommendations on using feature flags to manage new unreleased features in our continuous integration environment.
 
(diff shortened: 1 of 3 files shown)


packages/documentation/src/pages/platform/tools/feature-flags.mdx
```diff
@@ -45,9 +43,80 @@ vagovdev: false
...
+If you want to make a smaller conditional logic change than creating a whole new file, you can import the featureFlags file and check the flags directly:
+
+```js
+const { featureFlags, enabledFeatureFlags } = require('./src/site/utilities/featureFlags');
+
+...
+
+if (enabledFeatureFlags[featureFlags.FEATURE1]) {
+  // do something when the flag is on
+}
```
ncksllvn: It might be nice to make `enabledFeatureFlags` a `Set` at some point.

jbalboni: I decided not to do that because I wasn't sure if it would play nice with Liquid. But yeah, that's something we could definitely try out.

PR #99, opened 2019-05-02
platform documentation search indexing

rianfowler:
This should cover the remaining pages available on the live site which will close out the associated ticket. Will need someone to address changes and merge this in since I'm out.


packages/documentation/src/pages/platform/front-end-standards/accessibility.mdx
```diff
@@ -1,3 +1,8 @@
+---
+title: Accessibility standards
+tags: '508'
```
rianfowler: forces 508 to be a string

PR #101, opened 2019-05-04
Bug for accordion with clicking on accordion text and missing aria-expanded attribute: Bug/18411

accbjt:
When implementing formation.js in vets-website we found that when clicking on the banner text it didn't trigger the accordion. Also usa-accordion-button with no aria-expanded attribute displays a - icon by default. It needs to display '+' icon.


packages/formation/js/accordion.js
```diff
@@ -15,6 +15,16 @@ const isElementInViewport = (
   );
 };
 
+const addAriasExpandedAttr = () => {
+  document.querySelectorAll('.usa-accordion-button').forEach(element => {
+    const hasAriasExpandedAttr = element.getAttribute('aria-expanded');
+
+    if (!hasAriasExpandedAttr) {
+      element.setAttribute('aria-expanded', false);
```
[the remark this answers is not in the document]
accbjt: This will be converted to a string value in html.

packages/formation/js/accordion.js
```diff
@@ -40,11 +50,12 @@ const addAccordionClickHandler = () => {
...
     .forEach(element => {
       element.addEventListener('click', e => {
-        const target = e.target;
+        const target = e.target.closest('.usa-accordion-button');
 
         // Checks whether the button has a click event already assigned to it.
+        // and if it is a .usa-accordion-button.
         // Specifically React Components.
-        if (!target.onclick) {
+        if (target && !target.onclick) {
```
ncksllvn: Should the `target` variable be renamed to `accordionButton`, since it's not the `event.target` anymore?

packages/formation/__tests__/js/accordion/accordion.test.js
```diff
@@ -103,6 +103,14 @@ document.body.innerHTML = `
 `;
 
 describe('accordion', () => {
+  it('should set aria-expanded attribute on load if non exists', () => {
+    require('../../../js');
```
ncksllvn: I haven't seen a `require` inside an `it()`. Does this work differently than just importing at the top of the file?

accbjt: @ncksllvn just checked and I thought I needed this to reload on each test but just loading it after the DOM was created worked. Thanks. I think I got the first test working then just copy pasted it and didn't realize I had it was still there.

packages/formation/js/accordion.js
accbjt: Yeah that makes sense.

PR #105, opened 2019-05-09
Fix the paths to team repo pages

jbalboni:
With the team repo re-org, some of the paths changed, which breaks some pages we're trying to pull in.


accbjt:
@jbalboni There are a couple things I notice but I'm not sure it needs to happen for this PR. The Sidebar component doesn't have anything in it. Does this want to have links? Also on the actual github pages there are links on the page that don't work due to not loading them in. Do we want to load these pages also so they are available within the site?

jbalboni:
Yeah, I think there's some work to actually integrate these pages, if we want them. I was mainly trying to juggle get the build working again. I will create a ticket for figuring out the right way to use these pages.

PR #108, opened 2019-05-13
modularize ombinfo

rianfowler:
- adds support for OMB info to render its children 
- make it backward compatible 

verified (yalc into [this branch](https://github.com/department-of-veterans-affairs/vets-website/pull/10061)): 

- shows new content 
   - **burials**: http://localhost:3001/burials-and-memorials/pre-need/form-10007-apply-for-eligibility/introduction
  - **education**: http://localhost:3001/education/submit-school-feedback/introduction
  - **healthcare**: http://localhost:3001/health-care/apply/application/introduction
- shows old content 
  - http://localhost:3001/education/apply-for-education-benefits/application/5490/introduction


packages/formation-react/src/components/OMBInfo/OMBInfo.jsx
```diff
@@ -86,7 +85,9 @@ class OMBInfo extends React.Component {
         </div>
         <Modal
           cssClass="va-modal-large"
-          contents={modalContents(resBurden)}
+          contents={
+            this.props.children ? this.props.children : modalContents(resBurden)
```
rianfowler: Decided against making OMBInfo handle a render prop since it's not really a data provider.

 This requires some repetition but this seems fine as long as they're function components: https://github.com/department-of-veterans-affairs/vets-website/pull/10061/files#r283542665

Rendering the children means the data provider (InroductionPage most of the time) needs to do some composition with the benefit of being able to test new content without updating the formation library. Decided this was better than forcing the OMBInfo component to deal with that concern (e.g. pass switch type for health/education/burial and have OMBInfo render only one of the set chunks of content). 

LMK what you think.

jbalboni: I think this is a good approach. Do we even want the privacy content components in formation-react? it's somewhat specific to our forms and unless there's a fully generic version, it seems like those components could live in vets-website.

rianfowler: yeah that works for me- I'll move them into `vets-website`

PR #106, opened 2019-05-10
Sidebar causes page to not scroll: 18507

accbjt:
Sidebar causes page to not scroll when open. This only happens when going from mobile to desktop.

Steps:
1. Go to a page that has the sidebar.
2. Go to mobile.
3. Click on `more in this section`
4. Adjust to desktop browser size.
5. The page should scroll

Here's a screenshot of what it should do:

PR #111, opened 2019-05-15
add fix for sidebar scrolling and multiple open accordions

accbjt:
This is the same fix for https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/106. I found a bug. This should fix the issue with going from mobile to desktop. I also found a bug regarding multiple sidebar accordions being open when they shouldn't be.

Bug:
Screen Shot 2019-05-14 at 7 31 21 PM

Fix:
Screen Shot 2019-05-14 at 7 38 55 PM

### DIFF
--- a/packages/formation/js/accordion.js
+++ b/packages/formation/js/accordion.js
@@ -26,18 +26,12 @@ const addAriasExpandedAttr = () => {
 };
 
 const addAriaHiddenAttr = () => {
-  document
-    .querySelectorAll('.usa-accordion-content:not([aria-hidden])')
-    .forEach(el => {
-      const buttonElement = document.querySelector(
-        `[aria-controls="${el.id}"]`,
-      );
-      const hiddenValue = !toBoolean(
-        buttonElement.getAttribute('aria-expanded'),
-      );
+  document.querySelectorAll('.usa-accordion-content').forEach(el => {
+    const buttonElement = document.querySelector(`[aria-controls="${el.id}"]`);
+    const hiddenValue = !toBoolean(buttonElement.getAttribute('aria-expanded'));
 
-      el.setAttribute('aria-hidden', hiddenValue);
-    });
+    el.setAttribute('aria-hidden', hiddenValue);
+  });
 };
 
 const getOtherButtons = (element, target) =>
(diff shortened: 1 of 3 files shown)


packages/formation/js/sidenav.js
```diff
@@ -126,18 +127,22 @@ class SideBarMenu {
     this.lastTabbableElement.removeEventListener('keydown', this.captureFocus);
   }
 
-  closeMenu() {
+  addCloseMenuListener() {
```
accbjt: needed to separate the event listener from the actual close menu function. This allows me to call the closeMenu without having to add a new event listener.

jaredcunha-usds:
This seems to work on most pages, and the sidenav is working as expected now. However, I'm finding this:

On the accordion page, nothing works:
Screen Recording 2019-05-15 at 10 23 AM

Then in the left nav on the site, the accordions only close, and only close when I click an accordion button that is NOT the open accordion. However, this is on design.va.gov, which means one of the updates already merged in is causing this bug.
Screen Recording 2019-05-15 at 10 27 AM

accbjt:
@jaredcunha-usds Let me check it out.

accbjt:
@jaredcunha-usds found the issues. The accordion page was getting an error due to a `.usa-accordion-content` `Panel 2` element not having an `id`

Screen Shot 2019-05-15 at 9 36 03 AM

The second bug in the Sidenav was caused by having nested `.usa-accordion`. Parent and child elements both had `.usa-accordion` causing the click event to be assigned twice. So it would run the click event twice when clicking on it.
Screen Shot 2019-05-15 at 10 11 51 AM

These are both fixed now. Can you try it out and see if it's working on your machine.

packages/formation/js/accordion.js
```diff
@@ -49,57 +70,61 @@ const addAccordionClickHandler = () => {
...
-          }
-
-          if (hasAriaControlsAttr) {
-            const dropDownElement = document.getElementById(
-              accordionButton.getAttribute('aria-controls'),
-            );
-            const accordionButtonExpandedAttr = toBoolean(
-              accordionButton.getAttribute('aria-expanded'),
+      const parents = getParents(element, '.usa-accordion');
+
```
accbjt: Add this to handle case when parent and child both have the `.usa-accordion`.

jaredcunha-usds:
LGTM

PR #114, opened 2019-05-16
Sidenav tests

accbjt:
These are the only tests I can think of for the sidebar without using an integrated test runner.

### DIFF
--- a/packages/formation/__tests__/js/sidenav/sidenav.test.js
+++ b/packages/formation/__tests__/js/sidenav/sidenav.test.js
@@ -143,6 +143,9 @@ describe('Sidenav', () => {
 
   it('should hide sidenav when clicking on close button', () => {
     global.window.innerWidth = '600px';
+    const mobileSideNavButton = document.querySelector(
+      '.va-btn-sidebarnav-trigger',
+    );
     const navElement = document.querySelector('nav');
 
     const closeButton = navElement.querySelector('.va-btn-close-icon');
@@ -152,5 +155,6 @@ describe('Sidenav', () => {
     expect(
       navElement.classList.value.includes('va-sidebarnav--opened'),
     ).toEqual(false);
+    expect(mobileSideNavButton.getAttribute('hidden')).toEqual(null);
   });
 });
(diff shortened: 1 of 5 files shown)


packages/formation/__tests__/js/sidenav/sidenav.test.js
```diff
@@ -0,0 +1,163 @@
...
+    <b>More in this section</b>
+    <svg xmlns="http://www.w3.org/2000/svg" width="444.819" height="444.819" viewBox="0 0 444.819 444.819"><path fill="#ffffff" d="M352.025 196.712L165.885 10.848C159.028 3.615 150.468 0 140.185 0s-18.84 3.62-25.696 10.848l-21.7 21.416c-7.045 7.043-10.567 15.604-10.567 25.692 0 9.897 3.52 18.56 10.566 25.98L231.544 222.41 92.785 361.168c-7.04 7.043-10.563 15.604-10.563 25.693 0 9.9 3.52 18.566 10.564 25.98l21.7 21.417c7.043 7.043 15.612 10.564 25.697 10.564 10.09 0 18.656-3.52 25.697-10.564L352.025 248.39c7.046-7.423 10.57-16.084 10.57-25.98.002-10.09-3.524-18.655-10.57-25.698z"></path></svg>
+  </span>
+</button>
+<!-- End mobile toggle button -->
+
+<div class="medium-screen:vads-u-display--flex">
+  <div class="medium-screen:vads-l-col--3">
+    <!-- Begin Sidenav -->
+    <nav class="va-sidebarnav vads-u-width--full" id="va-detailpage-sidebar">
```
jbalboni: Do we need all of the html inside here? These are unit tests and they're testing a fairly limited set of functionality that doesn't appear to care about most of the content of the sidebar. Seems like we could slim down the html in here to just the minimum we need for the tests.

accbjt: Yeah, I did a little copy pasta here so I will fix it.

accbjt:
@jbalboni okay I removed some html from the tests.

jbalboni:
@accbjt Did you forget to push? I don't see any new commits.

accbjt:
@jbalboni Yes I did :). Okay it should be good now.

PR #116, opened 2019-05-20
17386 writing an e2e test

rianfowler:


packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
```diff
@@ -0,0 +1,153 @@
...
+- separate navigation from field input
+  - use a **main test file** for navigation, assertions, and calls helpers
+  - use a **helper file** that for filling out forms
+- create separate, numbered **main test files** to organize tests by their focus
+  - **00-all-fields.e2e.spec.js** - required and optional fields
+  - **01-required.e2e.spec.js** - only required fields
+  - **02-accessibility.e2e.spec.js** - validates accessibility
+  - **03-auth.e2e.spec.js** - validates authentication
+  - **04-cross-cutting-feature.e2e.spec.js** - validates one feature used across several pages (e.g. save in progress)
+- export the end to end test using `E2eHelpers.createE2eTest()`- see [Helpers](#helpers)
```
rianfowler: not sure what's going on here- LMK if you have any ideas otherwise I'll dig into it. 
Screen Shot 2019-05-20 at 12 03 40

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
```diff
@@ -0,0 +1,153 @@
...
+  - mock server is started as part of the end to end testing script
+- mock all api calls in your application e.g. submit and upload calls
+
+## Custom Nightwatch commands
+
+Nightwatch supports extending its client api with [custom commands](https://nightwatchjs.org/guide/#writing-custom-commands). Custom commands are located in `src/platform/testing/e2e/nightwatch-commands`
+
+_Command commands are available on the Nightwatch client e.g. `client.axeCheck()`. Below are some of the commonly used custom Nightwatch commands._
+
+- [`axeCheck(selector)`](https://github.com/department-of-veterans-affairs/vets-website/blob/d721cf0581c6b7d6f26903540dfc27f26a16a7be/src/platform/testing/e2e/nightwatch-commands/axeCheck.js#L4-L7) - uses the [`axe-core`](https://github.com/dequelabs/axe-core/blob/master/doc/developer-guide.md) library to run a series of tests to check for accessibility of content and functionality for DOM nodes in the selector
```
rianfowler: only included in this list if the command or helper is used more than a couple of times.

packages/documentation/src/sidebar.js
```diff
@@ -77,8 +77,8 @@ module.exports = {
               href: '/getting-started/common-tasks/new-unit-test',
             },
             {
-              name: 'Writing an end-to-end test',
-              // href: '/getting-started/common-tasks/write-e2e-test'
+              name: 'Write an end-to-end test',
```
rianfowler: gerund or verb?

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
```diff
@@ -0,0 +1,153 @@
+---
+title: Write an end to end test
+tags: Nightwatch, mockapi, e2e
```
rianfowler: review for completeness

erikphansen:
@tlei123 you might want to review these new docs since you're doing some e2e test work right now.

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
erikphansen: what if you place an empty line right above this bullet?

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
```diff
@@ -0,0 +1,153 @@
...
+---
+
+# Write an end to end test
+
+Front end engineers use end-to-end tests in `vets-website` to validate multipage applications with client-side routing. They are primarily used to assert that:
```
erikphansen: Since we also have some e2e test coverage for non-form apps perhaps the first sentence should be a little more flexible here? Also, this is minor, but maybe add `(e2e)` after the first mention of `end-to-end` here for clarity in case someone have never seen "e2e" before :)

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
```diff
@@ -0,0 +1,153 @@
...
+      data.veteranFullName.first,
+    )
+    .fill('input[name="root_veteranFullName_last"]', data.veteranFullName.last)
+}
+```
+_These are recommendations not requirements._
+
+- separate navigation from field input
+  - use a **main test file** for navigation, assertions, and calls helpers
+  - use a **helper file** that for filling out forms
```
erikphansen: ~~that~~

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
```diff
@@ -0,0 +1,153 @@
...
+- create separate, numbered **main test files** to organize tests by their focus
```
erikphansen: Add a `:` to the end of this line?

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
```diff
@@ -0,0 +1,153 @@
...
+  - **02-accessibility.e2e.spec.js** - validates accessibility
+  - **03-auth.e2e.spec.js** - validates authentication
+  - **04-cross-cutting-feature.e2e.spec.js** - validates one feature used across several pages (e.g. save in progress)
+- export the end to end test using `E2eHelpers.createE2eTest()`- see [Helpers](#helpers)
+- group tests by pages and use a comment to indicate what page is being tested
+- mock all api responses before starting the test. See [Mocking API responses](#mockimg-api-responses)
+- use `waitForElementVisible` before interacting with any element on the page
+- use `Timeouts` constants for setting timeouts (`platform/testing/e2e/timeouts.js`)
+- use helpers for filling data and performing actions on the page
+- perform `axeCheck` on the main body of the application on each page- see [Custom Nightwatch commands](#custom-nightwatch-commands)
```
erikphansen: `page-` -> `page -`

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
```diff
@@ -0,0 +1,153 @@
...
+
+**my-application-helpers.js**
+
+A mock server runs with the end to end tests to allow tests to make production-like calls.
+
+```
+const mock = require('platform/testing/e2e/mock-helpers');
+
+function initApplicationSubmitMock() {
+  mock(null, {
```
erikphansen: Maybe out of scope for these docs, but sometimes we need to pass a user token as the first arg to `mock()`. Also, even more likely out of scope, I've seen some examples of calling `client.mockData()` to mock a failure response for an API call, which is something the `mock()` helper doesn't appear to support.

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
```diff
@@ -0,0 +1,153 @@
...
+- [`selectYesNo(fieldName, predicate)`](https://github.com/department-of-veterans-affairs/vets-website/blob/d721cf0581c6b7d6f26903540dfc27f26a16a7be/src/platform/testing/e2e/nightwatch-commands/selectYesNo.js#L2-L6) - selects the provided options on yesNo widget at the field name when the predicate returns true
+
+## Helpers
+
+`src/platform/testing/e2e` contains other useful helpers
+
+- [`createE2eTest(beginApplication)`](https://github.com/department-of-veterans-affairs/vets-website/blob/d721cf0581c6b7d6f26903540dfc27f26a16a7be/src/platform/testing/e2e/helpers.js#L78-L81) - disables smooth scrolling and starts and ends test
+- [`overrideSmoothFormsScrolling(client)` and `overrideFormsScrolling(client)](https://github.com/department-of-veterans-affairs/vets-website/blob/d721cf0581c6b7d6f26903540dfc27f26a16a7be/src/platform/testing/e2e/nightwatch-commands/selectYesNo.js#L2-L6) - disables smooth scrolling during end to end testing
```
erikphansen: `overrideFormsScrolling(client)` is missing the closing backtick

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
rianfowler: I think I might just ignore it and see how it publishes. If it continues to be a problem I'll make a bug ticket.

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
rianfowler: end-to-end is good- I prefer not to use abbreviations in documentation but I'll add the the abbreviation in parenthesis.

Can you give me an example of an e2e test that doesn't focus on validating navigation? I didn't look very hard but most of the assertions are page navs. I think we usually rely on unit tests to validate the field level behavior.

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
rianfowler: It looks like the token is just added to the 2nd argument- @jbalboni is there a reason why we have this first argument and not just have people add it as `token:` property to the response object?

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
erikphansen: emptyPreferences.e2e.spec.js and selectedPreferences.e2e.spec.js come to mind

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
erikphansen: I don't think so. Here is an example of the token getting passed to the `mock()` call:

https://github.com/department-of-veterans-affairs/vets-website/blob/2cb102ae260833573905e033b67d93939254da4b/src/applications/hca/tests/hca-helpers.js#L380-L392

And that function gets called with a token here: https://github.com/department-of-veterans-affairs/vets-website/blob/26f7edca07ee0b40fb7bc3df29c46ed5f2ffee50/src/platform/forms/tests/save-in-progress/01-sip-autosave.e2e.spec.js#L9

jbalboni: Yeah, I don't see why it's written this way.

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
```diff
@@ -0,0 +1,153 @@
...
+- client applications render their inputs
+- client-side navigation occurs when the required fields are populated
+
+## end-to-end testing overview
+
+- `vets-website` uses [Nightwatch](https://nightwatchjs.org) to run the tests and provide the browser client
+- end-to-end tests are **collocated in application folder** with the application they test
+- Two node apps run with the end-to-end tests:
+  - `mockapi.js` - hosts mock responses (see [Mocking API responses](#mocking-api-responses))
+  - `test-server.js` - builds and hosts `vets-website` with support for client-side routes
```
jbalboni: I don't think I would say this "builds" vets-website, but it does run a local webserver that handles client-side routes, like you say. You have to build the site yourself if you're not running `yarn watch`, otherwise the e2e tests won't work.

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
```diff
@@ -0,0 +1,153 @@
...
+    )
+    .fill('input[name="root_veteranFullName_last"]', data.veteranFullName.last)
+}
+```
+_These are recommendations not requirements._
+
+- separate navigation from field input
+  - use a **main test file** for navigation, assertions, and calls helpers
+  - use a **helper file** for filling out forms
+- create separate, numbered **main test files** to organize tests by their focus:
```
jbalboni: I don't we're this intentional, I think the current naming is held over from previous ways of organizing tests. We might want to leave out anything but very basic guidelines until we have a coherent suggestion for naming.

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
jbalboni: Also, I wonder if we should suggest using the `client.mockData` helper instead of using `mock` directly? The interface is less confusing (token is an optional second param) and it will actually wait for the mock http call to finish before moving on with the test.

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
```diff
@@ -0,0 +1,153 @@
...
+- use `Timeouts` constants for setting timeouts (`platform/testing/e2e/timeouts.js`)
+- use helpers for filling data and performing actions on the page
+- perform `axeCheck` on the main body of the application on each page - see [Custom Nightwatch commands](#custom-nightwatch-commands)
+- assert that each navigation is successful
+
+## Mocking API responses
+
+**my-application-helpers.js**
+
+A mock server runs with the end-to-end tests to allow tests to make production-like calls.
```
jbalboni: We may want to mention somewhere that you should kill vets-api if you're running it locally, when you want to run an e2e test.

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
jbalboni: Maybe add `axe`

packages/documentation/src/pages/getting-started/common-tasks/new-end-to-end-test.mdx
```diff
@@ -0,0 +1,153 @@
...
+  - mock server is started as part of the end-to-end testing script
+- mock all api calls in your application e.g. submit and upload calls
+
+## Custom Nightwatch commands
+
+Nightwatch supports extending its client api with [custom commands](https://nightwatchjs.org/guide/#writing-custom-commands). Custom commands are located in `src/platform/testing/e2e/nightwatch-commands`
+
+_Command commands are available on the Nightwatch client e.g. `client.axeCheck()`. Below are some of the commonly used custom Nightwatch commands._
+
+- [`axeCheck(selector)`](https://github.com/department-of-veterans-affairs/vets-website/blob/d721cf0581c6b7d6f26903540dfc27f26a16a7be/src/platform/testing/e2e/nightwatch-commands/axeCheck.js#L4-L7) - uses the [`axe-core`](https://github.com/dequelabs/axe-core/blob/master/doc/developer-guide.md) library to run a series of tests to check for accessibility of content and functionality for DOM nodes in the selector
```
jbalboni: Should we link to the current master version of these files, or specific commit versions, like we do here? I tend to do the former, but it is possible for those links to break if the file is moved. On the other hand, it will link you to the most up to date version.

rianfowler: I think we'll link to the jsdocs whenever we start generating them. Most of these files don't change much.

PR #117, opened 2019-05-20
Object doesn't support property or method 'forEach' in Accordion.js

accbjt:


packages/formation/js/accordion.js
```diff
@@ -67,65 +74,65 @@ const getOtherButtons = (element, target) =>
...
-          const accordionButton = e.target.closest('.usa-accordion-button');
-
-          // Checks whether the button has a click event already assigned to it.
-          // and if it is a .usa-accordion-button.
-          // Specifically React Components.
-          if (accordionButton && !accordionButton.onclick) {
-            const multiSelectable = toBoolean(
-              element.getAttribute('aria-multiselectable'),
-            );
+  Array.from(
```
jbalboni: The diff went crazy here, I just added an `Array.from` and formatted with Prettier.

packages/formation/js/accordion.js
```diff
@@ -16,42 +20,45 @@ const isElementInViewport = (
...
+        );
+
+        el.setAttribute('aria-hidden', hiddenValue);
+      }
+    },
+  );
 };
 
-const getParents = (elem, selector) => {
+const getAccordionParents = elem => {
```
jbalboni: I updated this to avoid having to polyfill `matches`, though now that I've polyfilled `closest`, we could switch back.

PR #118, opened 2019-05-21
Create downtime-notifications.mdx

ncksllvn:
Migrated from https://github.com/department-of-veterans-affairs/vets-website/pull/10079

Ticket - https://github.com/department-of-veterans-affairs/vets.gov-team/issues/18612


packages/documentation/src/pages/platform/tools/downtime-notifications.mdx
```diff
@@ -0,0 +1,94 @@
+---
+title: Va.gov downtime notifications
+tags: pagerduty,maintenance windows,scheduled downtime
```
jbalboni: `Banner`, maybe?

packages/documentation/src/pages/platform/tools/downtime-notifications.mdx
```diff
@@ -0,0 +1,94 @@
...
+  fetchFromFancyService,
+};
+
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps,
+)(MyApp);
+```
+
+The render-flow for this container is:
```
jbalboni: I think you need a line break here. The list doesn't seem to render correctly.

jbalboni:
@ncksllvn You should also add it to the sidebar (`packages/documentation/src/sidebar.js`)

packages/documentation/src/pages/platform/tools/downtime-notifications.mdx
```diff
@@ -0,0 +1,94 @@
+---
+title: Va.gov downtime notifications
+tags: pagerduty,maintenance windows,scheduled downtime
+---
+
+# Downtime Notifications
+Downtime Notications are a mechanism for handling outages of external services in the consuming front-end applications. 
+
+## Overview
+- Available as a React component in [`vets-website`](https://github.com/department-of-veterans-affairs/vets-website/blob/cb1d8072a94b8847e0af0c697ff879b42db354f7/src/platform/monitoring/DowntimeNotification/index.jsx).
```
jbalboni: Link text surrounded by back ticks looks kinda weird on the site, I've been leaving out the back ticks if it's a link.

jbalboni: I asked this on another PR, I'm not sure if we want to link to the current master version of files or a specific commit like this. I typically do the former, but open to other opinions. We should probably be consistent about it.

packages/documentation/src/sidebar.js
```diff
@@ -267,6 +267,10 @@ module.exports = {
               name: 'Feature flags',
               href: '/platform/tools/feature-flags',
             },
+            {
+              name: 'Downtime notifications',
+              href: '/platform/tools/downtime-notifications',
+            },
```
ncksllvn: This made sense to me here, but LMK if there's a better fit somewhere else.

jbalboni: Yep, this is a good spot.

packages/documentation/src/pages/platform/tools/downtime-notifications.mdx
ncksllvn: I changed it to just link directly to master because I think it makes sense here, but I like linking to specific commits because we don't have to worry about the file moving or a line of code moving if we're linking to a specific line number.

packages/documentation/src/pages/platform/tools/downtime-notifications.mdx
```diff
@@ -0,0 +1,95 @@
+---
+title: Va.gov downtime notifications
+tags: pagerduty,maintenance windows,scheduled downtime,banner
```
ncksllvn: @jbalboni @rianfowler  Should all of these tags pull up `downtime notifications` in the search bar? It looks like `maintenance`, `scheduled`, and `banner` don't pull it up.

rianfowler: Didn't dive into it but I think the search tool we're using might be looking for `, ` when breaking up phrases for indexing. searching for "window" works but not "maintenance"

rianfowler: Could also be a bug with the display logic.

PR #119, opened 2019-05-22
The Pagination component is writing "undefined" in the next/previous page aria-labels

accbjt:
You can see this in the search app, or any other place that uses Pagination but does not use the ariaLabelSuffix prop. The component needs to print an empty string, not undefined, when that prop is not provided.

### DIFF
--- a/packages/formation-react/src/components/Pagination/Pagination.jsx
+++ b/packages/formation-react/src/components/Pagination/Pagination.jsx
@@ -13,16 +13,12 @@ class Pagination extends React.Component {
     this.pageNumbers = this.pageNumbers.bind(this);
   }
 
-  getAriaLabelSuffix() {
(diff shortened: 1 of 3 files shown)


packages/formation-react/src/components/Pagination/Pagination.jsx
```diff
@@ -13,12 +13,16 @@ class Pagination extends React.Component {
     this.pageNumbers = this.pageNumbers.bind(this);
   }
 
+  getAriaLabelSuffix() {
+    return this.props.ariaLabelSuffix || '';
```
jbalboni: Would setting a default prop for `ariaLabelSuffix` work? That might be slightly simpler.

jbalboni: Should probably bump the package version as well.

PR #124, opened 2019-05-23
add h2 support in feature component

jaredcunha-usds:
This PR will add support for styling `h2`s in featured content boxes. The styling simply removes the top margin. 

This should be included so that the proper semantic heading level can be applied.

Before:
Screen Shot 2019-05-23 at 11 54 11 AM

After:
Screen Shot 2019-05-23 at 11 54 20 AM

### DIFF
--- a/packages/formation/sass/base/_va.scss
+++ b/packages/formation/sass/base/_va.scss
@@ -400,6 +400,10 @@ article > h1 {
     padding-bottom: 0.5em;
   }
 
+  h2 {
+    margin-top: 0;
+  }
+
   h3 {
     margin-top: 0;
   }
--- a/packages/documentation/package.json
+++ b/packages/documentation/package.json
@@ -4,7 +4,7 @@
   "description": "Developer Document Site for VA.gov",
   "version": "0.1.0",
   "dependencies": {
-    "@department-of-veterans-affairs/formation": "^6.3.3",
+    "@department-of-veterans-affairs/formation": "^6.3.7",
     "@fortawesome/fontawesome-free": "^5.8.1",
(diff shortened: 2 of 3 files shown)


packages/formation/sass/base/_va.scss
```diff
@@ -400,6 +400,10 @@ article > h1 {
     padding-bottom: 0.5em;
   }
 
+  h2 {
```
jbalboni: This is a pretty broad change that I'd want to look at applied to vets-website before we merge it here. I think it may affect some paragraph spacing in content pages.

jaredcunha-usds: I ran a search on `vagov-content` and looked at 12 pages of resuluts to see if any h2s are in use in a `feature` component. That does not appear to be the case. Mostly, I'm seeing h3 as the top of level heading inside a `feature`, which may not always be semantically correct. On some of the pages, the headings skip from h1 -> h3.

jaredcunha-usds: wonder if @1Copenut might have thoughts on this.

jbalboni: Oh, I actually misread the code here. Didn't realize it was nested. (_shakes fist at sass nesting_)

jaredcunha-usds: _also shakes fist at sass nesting_

jenniferlee-dsva:
Thanks all, especially @jaredcunha-usds  and @rtwell  who correctly identified this as a code/not font source. 

RE ^ comment - most featured box headers being h3: https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/124#discussion_r287037224  That's true now, but the content team is working on correcting these to match the semantic hierarchy of headers for CMS migration.

PR #126, opened 2019-05-23
Add option to toggle borders in CollapsiblePanel

narin:
Add a `borderless` propType to control whether the content panels have a border or not.


narin:
Thanks for the feedback @accbjt.  I removed the defaultProps.  Please take another look when you get a chance.

PR #132, opened 2019-06-05
[Loading Indicator] Switch from SVG to CSS

ncksllvn:
## Description
Currently, the loading-indicator uses an animated SVG as its spinner. This isn't friendly with Internet Explorer or Edge - in those browsers, it's just a blue outline of a circle. This wasn't an issue before because most of the time the spinner was only shown for a split second. However, the new Community Cares functionality can take a full minute to load, and the broken SVG makes it appears that much more broken.

This PR proposes we switch to a CSS spinner. [transition](https://caniuse.com/#feat=css-transitions) appears to be supported by all of the browsers VA.gov supports.

Ticket - https://github.com/department-of-veterans-affairs/vets.gov-team/issues/18818

## Testing done
Tested UX of Facility Locator
- Tested app booting up
- Tested queries

## Screenshots

ezgif-4-a6dea2f77764

ezgif-4-5139dc256428

## Acceptance criteria
- [x] Spinner works in IE/Edge
- [x] HTML Is not changed

## Definition of done

### DIFF
--- a/packages/formation/sass/modules/_m-loading-indicator.scss
+++ b/packages/formation/sass/modules/_m-loading-indicator.scss
@@ -15,7 +15,7 @@
       border-radius: 50%;
       border: 8px solid #0071BB;
       border-color: #0071BB transparent #0071BB transparent;
-      animation: lds-dual-ring 1.2s linear infinite;
+      animation: spin 1.2s linear infinite;
     }
 
     &:focus {
@@ -28,7 +28,7 @@
   padding-bottom: 1em;
 }
 
-@keyframes lds-dual-ring {
+@keyframes spin {
   0% {
     transform: rotate(0deg);
   }
--- a/packages/formation/sass/modules/_m-loading-indicator.scss
+++ b/packages/formation/sass/modules/_m-loading-indicator.scss
@@ -3,18 +3,18 @@
 
   .loading-indicator {
     display: block;
-    height: 8em;
-    margin-top: 2em;
+    height: units(8);
+    margin-top: units(2);
 
     &:after {
       content: " ";
       display: inline-block;
-      width: 7em;
-      height: 7em;
+      width: units(7);
+      height: units(7);
       margin: 1px;
       border-radius: 50%;
-      border: 8px solid #0071BB;
-      border-color: #0071BB transparent #0071BB transparent;
+      border: 8px solid  $color-primary;
+      border-color:  $color-primary transparent  $color-primary transparent;
       animation: spin 1.2s linear infinite;
     }
 
@@ -25,7 +25,7 @@
 }
 
 .async-loader {
-  padding-bottom: 1em;
+  padding-bottom: units(1);
 }
 
 @keyframes spin {
--- a/packages/formation/sass/modules/_m-loading-indicator.scss
+++ b/packages/formation/sass/modules/_m-loading-indicator.scss
@@ -2,10 +2,22 @@
   text-align: center;
 
   .loading-indicator {
-    min-height: 8em;
-    background: url('#{$formation-image-path}/loading-state.svg') center center no-repeat;
-    background-size: contain;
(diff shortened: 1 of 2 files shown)


packages/formation/sass/modules/_m-loading-indicator.scss
```diff
@@ -2,10 +2,22 @@
...
+    margin-top: 2em;
+
+    &:after {
+      content: " ";
+      display: inline-block;
+      width: 7em;
+      height: 7em;
+      margin: 1px;
+      border-radius: 50%;
+      border: 8px solid #0071BB;
```
jaredcunha-usds: Let's use our color tokens instead of hex codes here: `$color-primary` is the token for `#0071BB`.

packages/formation/sass/modules/_m-loading-indicator.scss
```diff
@@ -2,10 +2,22 @@
...
-    background-size: contain;
-    text-align: center;
+    display: block;
+    height: 8em;
+    margin-top: 2em;
+
+    &:after {
+      content: " ";
+      display: inline-block;
+      width: 7em;
```
jaredcunha-usds: Using `em` units, the size of the loader could change if the font-size is set on the container.  Can we look at using the spacing units `units` function for height & width (works here since this isn't a super huge thing), margin, and padding? https://design.va.gov/design/spacing-units

jaredcunha-usds:
@ncksllvn This is awesome. I still want to test it out locally when I get in. In the meantime, I have made a couple of minor suggestions. I'd considering adding to the AC that HTML of the existing loading indicator not change, and it appears from what I can tell in the CSS, that is met.

One thing I really like about the new loading now is that because it's pure CSS, we can add modifiers:
- Small indicators
- Indicators of different colors to use on different backgrounds.

packages/formation/sass/modules/_m-loading-indicator.scss
```diff
@@ -15,3 +27,12 @@
 .async-loader {
   padding-bottom: 1em;
 }
+
+@keyframes spin {
```
jaredcunha-usds: Thoughts on moving this to a `_keyframes.scss` file?

packages/formation/sass/modules/_m-loading-indicator.scss
ncksllvn: 🤦‍♂ I must have been really tired when I wrote this!

packages/formation/sass/modules/_m-loading-indicator.scss
ncksllvn: Hm, I see it either way, but for consistency across the design system it's probably better in `_keyframes.scss`.

ncksllvn: Hm, actually, I thought we already had a keyframes file, but since we don't it might be better to do this separately. What do you think?

ncksllvn:
> I'd considering adding to the AC that HTML of the existing loading indicator not change, and it appears from what I can tell in the CSS, that is met.

Sorry, what does AC stand for? 😆

ncksllvn:
I think this is ready for another look, btw. Trying to get it published today if possible, because a lot of eyes are on the Community Cares section of the Facility Locator today!

jaredcunha-usds:
@ncksllvn AC = Acceptance Criteria. We government folks will make anything an acronym.

ncksllvn:
@jaredcunha-usds ah I should have known that but I couldn't get `air conditioner` out of my head 😆 Added to the AC!

ncksllvn:
This can be improved a lot - probably the look/feel of the spinner and definitely with @jaredcunha-usds's suggestions - but I think it's okay as a starting point. Gonna go ahead and merge...

jaredcunha-usds:
Just tested. Looks good.

PR #133, opened 2019-06-06
fix lead line height

jaredcunha-usds:
## Description
This is a minor fix to the lead text `line-height` suggested by @rtwell.

What was happening is that the  `line-height` was computing to `30px` for this:
```
<div class="va-introtext">
   <p>Some text</p>
</div>
```

But it would be `34px` for this:
```
<p class="va-introtext">Some text</p>
```

We want the `line-height` to be `34px`.  Looks like the paragraph's base styles were overriding the CSS in `va-introtext`, so this patch just reinforces that with some extra specificity.

### DIFF
--- a/packages/documentation/package.json
+++ b/packages/documentation/package.json
@@ -4,7 +4,7 @@
   "description": "Developer Document Site for VA.gov",
   "version": "0.1.0",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


rtwell:
❤️

PR #134, opened 2019-06-07
Add release process for formation and formation-react

accbjt:
## Description
Should walk through the steps for making the change, updating versions in vfs tools and vets-website.
&
Integrate Formation releases with the Github release tab

## Testing done
local testing

## Acceptance criteria
- [ ] Should have the detailed process on how to update formation and formation-react
- [ ] Running `yarn release` in the package folder should generate a release with the prefix of the package and version number (formation-v6.2.1).
- [ ] GH release info should have the changes listed in that release (https://github.com/usds/us-forms-system/releases is a good format example, as well as other open source projects)

## Definition of done


packages/documentation/src/pages/getting-started/common-tasks/updating-formation.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+You'll need to choose what type of version update to make:
+
+- `patch` - for bug fixes and minor changes
+- `minor` - for new features that don't break current features or require changes in consuming applications
+- `major` - for backwards breaking changes
+
+> If you are unsure of what to pick, do a major version update on modules that are changing.
+
+### 2. Create a release
```
jbalboni: Shouldn't this step be after step 4?

packages/documentation/src/pages/getting-started/common-tasks/updating-formation.mdx
```diff
@@ -0,0 +1,55 @@
+---
+title: Making Updating Formation and Formataion React
```
jbalboni: `Formataion`

Also, I think you meant to replacing "Making" with "Updating"?

packages/documentation/src/pages/getting-started/common-tasks/updating-formation.mdx
```diff
@@ -0,0 +1,55 @@
...
+tags: formation, formation-react, update, release, publish, npm
+---
+
+## Making updates to formation
+You can find a detailed guide on contributing to formation at [https://dev-design.va.gov/documentation/contributing-to-formation](https://dev-design.va.gov/documentation/contributing-to-formation)
+
+## Making updates to formation-react
+1. Components should be created and tested in vets-website first before being moved over to formation-react
+2. Transfer component to formation-react
+3. Test new node_module component locally. [Check out these instructions on how to do so](/previewing-changes)
```
jbalboni: I'm not sure "node_module component" makes sense. Maybe say "Test the formation-react version of the component in vets-website before publishing"

scripts/release.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable camelcase */
+/* eslint-disable no-console */
+import fetch from 'node-fetch';
+
+const version = process.env.VERSION;
+
+const createTag = async () => {
+  try {
+    const payload = {
+      tag_name: `${process.env.PACKAGE}-v${version}`,
```
jbalboni: Does this create a tag?

packages/documentation/src/pages/getting-started/common-tasks/updating-formation.mdx
```diff
@@ -0,0 +1,55 @@
+---
```
jbalboni: It looks like this doc replaces the content in the README. Can we update the README to point to this page for how to make changes to formation/formation-react, so we don't have duplicate, out of date content in there?

scripts/release.js
```diff
@@ -0,0 +1,38 @@
...
+      name: `${process.env.PACKAGE}-v${version}`,
+      body: `${process.env.PACKAGE} release version v${version}`,
+      draft: false,
+      prerelease: false,
+    };
+
+    const response = await fetch(
```
jbalboni: I'm assuming we can iterate on this api call and include a zipped version of the formation build output.

packages/documentation/src/pages/getting-started/common-tasks/updating-formation.mdx
accbjt: @jbalboni according to GitHub flow process. The release happens right before merging to master. Can you confirm if this is correct?

scripts/release.js
accbjt: Yes the request creates a tag and a release. I initially had some code to create a tag through the shell but this actually does both. Also found out that we can actually prefix the tags.

scripts/release.js
accbjt: We should be able to. I have some time today to try this out and add a function that can actually do this for us. Here's a link that describes how to do it.

https://developer.github.com/v3/repos/releases/#upload-a-release-asset

packages/documentation/src/pages/getting-started/common-tasks/updating-formation.mdx
jbalboni: I don't think GH flow matches what we've been doing exactly. We're similar, but we always create releases from code that is in master, but here and in vets-website.

accbjt: Okay, I will update this to reflect what we are already doing.

jaredcunha-usds:
I'm not 100% sure what this is all doing, but I'd love to see a walkthrough when it's ready!

scripts/release.js
accbjt: Looks like this will take longer to implement, but I do think this is possible. The only gotcha is zipping the files. Since formation and formation-react builds are different you will have to indicate what files are needed in the zip folder. I think formation is pretty straight forward but formation-react needs to be more defined. Overall I think this is possible and we shouldn't have a problem implementing this in the future.

accbjt:
@jaredcunha-usds It's pretty simple. You run `npm run release` inside your package folder and it will generate a tag and release for you in GitHub.

accbjt:
@jbalboni I made the changes you suggested. The is a build error on Jenkins. It seems like a module that gatsby is using isn't installing correctly. Can you take a look at this for me. Thanks

accbjt:
This PR is also connected to this ticket.

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15407

jaredcunha-usds:
> @jaredcunha-usds It's pretty simple. You run npm run release inside your package folder and it will generate a tag and release for you in GitHub.

Oh, awesome!!!

PR #137, opened 2019-06-19
Implement PromoBanner

ncksllvn:
## Description
Formerly known in vets-website as an "announcement", this PR redesigns and implements a promo-banner component. 

There are three types of promo-banners, each with their own FA icon -

1. Announcement (this is why I changed the name)
2. News
3. Email Signup

Ticket - https://github.com/department-of-veterans-affairs/vets.gov-team/issues/18175

Once this is ready, I'll update the vets-website PR to begin consuming the new `PromoBanner` component, https://github.com/department-of-veterans-affairs/vets-website/pull/10269.

## Testing done
- Locally tested in vets-website

#### Large screen
image

#### Small screen
image

#### Large screen
image

#### Small Screen
image

### News example
image

### Email Signup Example
image

## Acceptance criteria
- [x] Component is implemented and up to standards

## Definition of done

### DIFF
--- a/packages/formation-react/src/components/PromoBanner/PromoBanner.jsx
+++ b/packages/formation-react/src/components/PromoBanner/PromoBanner.jsx
@@ -24,40 +24,37 @@ class PromoBanner extends React.Component {
 
     return (
       <div className="vads-c-promo-banner">
-        <div className="vads-l-grid-container">
-          <div className="vads-c-promo-banner__body">
-            <div className="vads-c-promo-banner__content">
-              <div className="vads-c-promo-banner__content-icon">
-                <span className="fa-stack fa-lg">
-                  <i className="vads-u-color--white fa fa-circle fa-stack-2x" />
-                  <i className={iconClasses} />
-                </span>
-              </div>
-
-              {this.props.render ? (
-                this.props.render()
-              ) : (
-                <a
-                  className="vads-c-promo-banner__content-link"
-                  href={this.props.href}
-                  onClick={this.props.onClose}
-                >
-                  {this.props.text}{' '}
-                  <i className="fas fa-angle-right vads-u-margin-left--1" />
-                </a>
-              )}
+        <div className="vads-c-promo-banner__body">
+          <div className="vads-c-promo-banner__content">
+            <div className="vads-c-promo-banner__content-icon">
+              <span className="fa-stack fa-lg">
+                <i className="vads-u-color--white fa fa-circle fa-stack-2x" />
+                <i className={iconClasses} />
+              </span>
             </div>
 
-            <div className="vads-c-promo-banner__close">
-              <button
-                type="button"
-                aria-label="Dismiss this announcement"
+            {this.props.render ? (
+              this.props.render()
+            ) : (
+              <a
+                className="vads-c-promo-banner__content-link"
+                href={this.props.href}
                 onClick={this.props.onClose}
-                className="va-button-link vads-u-margin-top--1"
               >
-                <i className="fas fa-times-circle vads-u-font-size--lg" />
-              </button>
-            </div>
+                {this.props.text} <i className="fas fa-angle-right" />
+              </a>
+            )}
+          </div>
+
+          <div className="vads-c-promo-banner__close">
+            <button
+              type="button"
+              aria-label="Dismiss this announcement"
+              onClick={this.props.onClose}
+              className="va-button-link vads-u-margin-top--1"
+            >
+              <i className="fas fa-times-circle vads-u-font-size--lg" />
+            </button>
           </div>
         </div>
       </div>
--- a/packages/formation-react/src/components/PromoBanner/PromoBanner.jsx
+++ b/packages/formation-react/src/components/PromoBanner/PromoBanner.jsx
@@ -23,11 +23,11 @@ class PromoBanner extends React.Component {
     );
 
     return (
-      <div className="va-promo-banner">
+      <div className="vads-c-promo-banner">
         <div className="usa-grid-full">
-          <div className="va-promo-banner-body">
-            <div className="va-promo-banner-content">
-              <div className="va-promo-banner-content-icon">
+          <div className="vads-c-promo-banner__body">
+            <div className="vads-c-promo-banner__content">
+              <div className="vads-c-promo-banner__content-icon">
                 <span className="fa-stack fa-lg">
                   <i className="vads-u-color--white fa fa-circle fa-stack-2x" />
                   <i className={iconClasses} />
@@ -38,7 +38,7 @@ class PromoBanner extends React.Component {
                 this.props.render()
               ) : (
                 <a
-                  className="va-promo-banner-content-link"
+                  className="vads-c-promo-banner__content-link"
                   href={this.props.href}
                   onClick={this.props.onClose}
                 >
@@ -48,7 +48,7 @@ class PromoBanner extends React.Component {
               )}
             </div>
 
-            <div className="va-promo-banner-close">
+            <div className="vads-c-promo-banner__close">
               <button
                 type="button"
                 aria-label="Dismiss this announcement"
--- a/packages/formation-react/src/components/PromoBanner/PromoBanner.jsx
+++ b/packages/formation-react/src/components/PromoBanner/PromoBanner.jsx
@@ -24,7 +24,7 @@ class PromoBanner extends React.Component {
 
     return (
       <div className="vads-c-promo-banner">
-        <div className="usa-grid-full">
+        <div className="vads-l-grid-container">
           <div className="vads-c-promo-banner__body">
             <div className="vads-c-promo-banner__content">
               <div className="vads-c-promo-banner__content-icon">
--- a/packages/formation-react/src/components/PromoBanner/PromoBanner.jsx
+++ b/packages/formation-react/src/components/PromoBanner/PromoBanner.jsx
@@ -23,7 +23,7 @@ class PromoBanner extends React.Component {
     );
 
     return (
-      <div className="vads-c-promo-banner">
+      <div className="vads-c-promo-banner vads-c-promo-banner--fixed-bottom">
         <div className="vads-c-promo-banner__body">
           <div className="vads-c-promo-banner__content">
             <div className="vads-c-promo-banner__content-icon">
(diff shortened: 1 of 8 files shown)


packages/formation/sass/modules/_m-promo-banner.scss
```diff
@@ -0,0 +1,54 @@
...
+  font-weight: 700;
+
+  &-body {
+    display: table;
+    width: 100%;
+  }
+
+  &-content, &-close {
+    display: table-cell;
+  }
```
ncksllvn: I used `display: table` because of the responsive behavior - on smaller screens, the close button's cell should span the whole height of the parent with the close button itself centered vertically and horizontally. I explored flexbox for awhile but there didn't seem to be an easy way to do this. 

image

jaredcunha-usds:
I had a couple thoughts:

1) I know it results in uglier Sass, but  would you be able use more searchable selectors: https://design.va.gov/documentation/developers#searchable-selectors

2) Let's start using the design system component naming convention: https://design.va.gov/documentation/naming
So that would be somethling like: `.vads-c-promo-banner`, `vads-c-promo-banner__body`, etc.

3) Did we ever land on a naming convention for this? I was thinking announcement, but maybe I'm wrong. @rtwell, do you remember?

4) Did you want to try one of the VA grid container instead of `usa-grid`? I'm not sure if you'd need the default for the `--full` modifier here. We're using that for the new page layouts. (https://design.va.gov/layout/flexbox-grid)

ncksllvn:
All sounds good, I'll try to make those updates before EOD.

On point 3, the consensus was "announcement", but there are 3 variants, one of which is "announcement". I thought this conflict was confusing, which made me think "announcement" was too vague. I like how "promo-banner" helps express its role and appearance.

jaredcunha-usds:
But I think that creates a little confusion to with the alert banner, and then the promo right rail thing. 

Is it bad practice to do:
```
const PROMO_BANNER_TYPES = {
  default: 'default', // instead of announcement
  news: 'news',
  emailSignup: 'email-signup',
};
```

jaredcunha-usds:
FWIW - I think that banner is the best most appropriate name. We don't use 'banner' anywhere in the full-width alerts that we refer to as a banner. 

So maybe, yeah, forget what I said. This can probably be the new banner.

rtwell:
a couple items (i'm not sure this is the time and place for this level of polish, but i'm getting thoughts down now):

1) tbh I don't remember where we landed for a name. we have so many announcement-like interactions with alerts and banners and notifications…I'm not sure what the best name should be. **Banner** seems to be the most fitting to me, but I defer to what makes sense to you all from a code perspective. 

2) can we shrink the size of those `angle-right` icons? they seem to be rendered too large. The type and the icon should be the same size.

3) our line height feels too generous. can we snug it up a bit? In sketch it's 16px type / 20px line-height if that helps. 

thank you for your work on getting this done!

jaredcunha-usds:
Banner it is!

ncksllvn:
Are we okay with calling it the _Promo_-Banner? Working with the content team, I hear plain "banner" used constantly, usually interchangeably with alert boxes.

jaredcunha-usds:
Yeah, I am. I thought the consensus was announcement because that's what we all agreed; but banner or promo banner is the best descriptive name. The alerts should not have two different names based on the width. So what the content team calls a banner should be a full width alert. @rtwell - what you think? We can work with Jen and the content team adopt the new-but-not-actually-new convention?

rtwell:
we should be more strict about our language, and get out of the habit of referring to our alerts as banners. we have promo spots currently, so I find that confusing to use promo for both.

rtwell:
we can def work with Jen and her team to bring more clarity to these components for sure!

jaredcunha-usds:
@rtwell - Agreed!

packages/formation/sass/modules/_m-promo-banner.scss
```diff
@@ -0,0 +1,54 @@
...
+  background-color: $color-gray-lightest;
+  color: $color-link-default;
+  font-weight: 700;
+
+  .vads-c-promo-banner__body {
+    display: table;
+    width: 100%;
+  }
+
+  .vads-c-promo-banner__content, va-promo-banner-close {
```
jaredcunha-usds: looks like a typo here BTW. 

Should be:
```
 .vads-c-promo-banner__content, 
.vads-c-promo-banner__close {
```

ncksllvn: Haha yep, you caught that right after I wondered why my close button didn't look right on small screens

ncksllvn:
Okay, I think I made all of the changes. Updated screenshots -


## Large
image

## Small
image

rtwell:
looking great per usual @ncksllvn !

can we skootch the angle-right closer to the type (to the left)? maybe about half the distance it is now? secondly, on the mobile banner, can we center (both vertically and horizontally) the close "X" in that square space? it looks like its down and to the left.

ncksllvn:
@jaredcunha-usds it looks like switching the grid class added some padding-right on small screens, causing a small issue with the close button -

image

rtwell:
I'd also pull all of the type closer to the left margin so the type left aligns with "Search" above it
image

packages/formation/sass/modules/_m-promo-banner.scss
```diff
@@ -0,0 +1,55 @@
...
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  width: 100%;
+
+  background-color: $color-gray-lightest;
+  color: $color-link-default;
+  font-weight: 700;
+
+  .vads-c-promo-banner__body {
```
erikphansen: Entirely optionally but I think it's a bit more SASS-y (heh heh) to write these nested classes with the `&` so we aren't rewriting the same class name. So `.vads-c-promo-banner__body` -> `&__body` I wonder if anyone else has strong feelings about this one way or another.

ncksllvn:
Updated -

## Large
image

## Small
image

packages/formation/sass/modules/_m-promo-banner.scss
jaredcunha-usds: @erikphansen It is, but it makes code a lot harder to search. I've run into lots of issues trying to find things myself. Which is why I added this rule: https://design.va.gov/documentation/developers#searchable-selectors

jaredcunha-usds:
Crap - @ncksllvn one more suggestion? 

Can you get away with not nesting your selectors to keep the specificity low?

packages/formation/sass/modules/_m-promo-banner.scss
jbalboni: @ncksllvn The VCL banner has that same behavior that with flexbox, I believe

jaredcunha-usds:
So instead of:

```
.vads-c-promo-banner {
  .vads-c-promo-banner__body {}
}
```

Do:
```
.vads-c-promo-banner {
}

.vads-c-promo-banner__body {
}
```

packages/formation/sass/modules/_m-promo-banner.scss
```diff
@@ -0,0 +1,57 @@
+.vads-c-promo-banner {
```
jbalboni: Is it worth separating the positioning here from the styling? Granted, we only have the one use case for it now, but the style of the banner is independent of where it's positioned on the screen and putting those styles in two separate classes might make it more flexible later.

Like `vads-c-promo-banner` for the visual styles and `vads-c-promo-banner--fixed-bottom` to fix its position to the bottom of the page. 

Curious what @jaredcunha-usds thinks.

ncksllvn: Oh hm, that makes sense to me. That's a pretty quick change...

jaredcunha-usds: I think we can leave the default styling be fixed at the bottom, so I don't think we need to create the variant. Maybe a variant for whatever changes in the future, but I don't think we'll need `vads-c-promo-banner--fixed-bottom`

ncksllvn: Kind of interesting seeing it without the `vads-c-promo-banner--fixed-bottom` applied -

image

ncksllvn: @jaredcunha-usds whoops! Should I revert the commit?

jaredcunha-usds: @ncksllvn Yeah, I think so. 😬

packages/formation-react/src/components/PromoBanner/PromoBanner.jsx
```diff
@@ -0,0 +1,76 @@
...
+  emailSignup: 'email-signup',
+};
+
+const PROMO_BANNER_ICONS = new Map([
+  [PROMO_BANNER_TYPES.announcement, 'fa-bullhorn'],
+  [PROMO_BANNER_TYPES.news, 'fa-newspaper'],
+  [PROMO_BANNER_TYPES.emailSignup, 'fa-envelope'],
+]);
+
+class PromoBanner extends React.Component {
```
jbalboni: This could probably be a function component.

packages/formation-react/src/components/PromoBanner/PromoBanner.jsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+const PROMO_BANNER_TYPES = {
```
jbalboni: Is there value in limiting the banner types? Or could we let people pass in the icon as a prop?

packages/formation-react/src/components/PromoBanner/PromoBanner.jsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+}
+
+PromoBanner.propTypes = {
+  type: PropTypes.oneOf(Object.values(PROMO_BANNER_TYPES)).isRequired,
+  onClose: PropTypes.func.isRequired,
+  render: PropTypes.func,
+  href: PropTypes.string,
+  text: PropTypes.string,
+  children: PropTypes.node,
```
jbalboni: Doesn't look like this is used?

packages/formation-react/src/components/PromoBanner/PromoBanner.jsx
ncksllvn: I considered having a `customIcon` prop, but the requirements seemed strict about having only three types of these so I decided to make the code reflect that. I'm fine with adding the `customIcon` prop though.

jbalboni: I don't have a strong opinion, if we're trying to limit the types, that totally makes sense.

PR #142, opened 2019-06-25
Add level prop to AlertBox

jbalboni:
## Description
This adds a level prop to the alert box, so that you can control the heading level directly and avoid axe errors related to skipping heading levels in an html page (e.g. following an `h1` with an `h3`, instead of an `h2`).

## Testing done
Unit testing

## Acceptance criteria
- [x] Component markup includes a `h` element that uses the level prop

## Definition of done

### DIFF
--- a/packages/formation-react/src/components/AlertBox/AlertBox.jsx
+++ b/packages/formation-react/src/components/AlertBox/AlertBox.jsx
@@ -73,6 +73,7 @@ class AlertBox extends React.Component {
   }
 }
 
+/* eslint-disable consistent-return */
 AlertBox.propTypes = {
   /**
    * Determines the color and icon of the alert box.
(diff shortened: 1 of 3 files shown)


packages/formation-react/src/components/AlertBox/AlertBox.jsx
```diff
@@ -126,12 +125,18 @@ AlertBox.propTypes = {
    * accented left edge or an icon
    */
   backgroundOnly: PropTypes.bool,
+
+  /**
+   * The header level to use with the headline prop
+   */
+  level: PropTypes.number,
```
erikphansen: my only suggestion would be to make the PropType for `level` a function that checks to see that the value is a number from 1-6.

PR #147, opened 2019-07-11
Add disabled state for buttons that are styled as links

ncksllvn:
## Description
This came up while working on disabled button-links on the disability calculator, mentioned in this [ticket](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19143).

Buttons that are styled to look like links should be gray when disabled.

## Testing done
I tested this on the [disability calculator](https://staging.va.gov/disability/about-disability-ratings-beta/).

## Screenshots
image

## Acceptance criteria
- [ ] Disabled button-links are styled correctly

## Definition of done

### DIFF
--- a/packages/formation/sass/base/_b-mixins.scss
+++ b/packages/formation/sass/base/_b-mixins.scss
@@ -154,7 +154,7 @@
   }
   &:disabled {
     text-decoration: none;
-    color: $color-gray !important;
+    color: $color-gray-light !important;
   }
 }
 
--- a/packages/formation/sass/base/_b-mixins.scss
+++ b/packages/formation/sass/base/_b-mixins.scss
@@ -152,6 +152,10 @@
     @include focus-gold-light-outline;
     outline-offset: 0;
   }
+  &:disabled {
+    text-decoration: none;
+    color: $color-gray !important;
+  }
 }
 
 @mixin modal-close-button {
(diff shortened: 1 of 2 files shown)


packages/formation/sass/base/_b-mixins.scss
```diff
@@ -152,6 +152,10 @@
     @include focus-gold-light-outline;
     outline-offset: 0;
   }
+  &:disabled {
+    text-decoration: none;
+    color: $color-gray-light !important;
```
ncksllvn: The override wasn't working without the `!important` here.

jaredcunha-usds:
So I guess this makes sense for diabled links when we need them, but I think @goldenmeanie has some thoughts about using disabled links vs not showing them at all. 

But going back to disabled states for links, we probably need to the cursor to be `default`, right? So we don't show the pointer?

ncksllvn:
@jaredcunha-usds It looks like it inherits that from USWDS, although I'm fine with adding the rule anyway if we want here as well -

image

jaredcunha-usds:
@ncksllvn If it already inherits from USWSD, no need for redundancy. I was unaware it was doing that.

ncksllvn:
@jaredcunha-usds I updated the description in the screenshot to help clarify exactly what this change does. I realized it was kinda unclear with a couple difference types of buttons.

PR #148, opened 2019-07-11
16156 - Add optional ariaLabelledBy prop to ErrorableCheckbox.

tlei123:
## Description
[16156](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/16156) - Add optional screen-reader support to read an external-heading instead of checkbox-label.

formation-react version bumped to 4.9.0 [minor update].

## Testing done
New unit-test added; All still passing (now 184 tests).

## Acceptance criteria
- [ ] Via Gatsby docs dev-server, bottom checkbox on page http://localhost:8000/visual-design/components/errorablecheckbox/ reads the external-heading above it.

## Definition of done

### DIFF
--- a/packages/formation-react/src/components/ErrorableCheckbox/ErrorableCheckbox.jsx
+++ b/packages/formation-react/src/components/ErrorableCheckbox/ErrorableCheckbox.jsx
@@ -1,6 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import _ from 'lodash';
+import { isUndefined, uniqueId } from 'lodash';
 
 class ErrorableCheckbox extends React.Component {
   constructor() {
@@ -9,7 +9,7 @@ class ErrorableCheckbox extends React.Component {
   }
 
   UNSAFE_componentWillMount() {
-    this.inputId = _.uniqueId('errorable-checkbox-');
+    this.inputId = uniqueId('errorable-checkbox-');
   }
 
   handleChange(domEvent) {
@@ -39,7 +39,7 @@ class ErrorableCheckbox extends React.Component {
     let className = `form-checkbox${
       this.props.errorMessage ? ' usa-input-error' : ''
     }`;
-    if (!_.isUndefined(this.props.className)) {
+    if (!isUndefined(this.props.className)) {
       className = `${className} ${this.props.className}`;
     }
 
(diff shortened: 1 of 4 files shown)


packages/formation-react/src/components/ErrorableCheckbox/ErrorableCheckbox.jsx
```diff
@@ -88,13 +89,27 @@ ErrorableCheckbox.propTypes = {
    */
   name: PropTypes.string,
   /**
-   * Label for the checkbox
+   * Label for the checkbox.
+   * IF you have a SEPARATE HEADING (outside this component) that “labels” this
+   * checkbox, do NOT pass {<span class="usa-sr-only">...</span>} here to just
+   * visually hide this label.
+   * Instead, pass an empty string ("") here, and pass that external heading‘s
```
jbalboni: Could we make this not a required prop? Forcing people to pass an empty string feels weird. We're also still rendering an empty label element, which maybe we don't want to do? Not sure how @1Copenut feels about that.

Or we could write a custom prop type here that says you need a label if you aren't using the aria labelled by prop.

tlei123: Custom-validators now added to label & ariaLabelledBy -- either one is required but not both.

tlei123:
@1Copenut , this PR is for updating the correct checkbox component we should be using for our benefit-cards on My-VA Find-Benefits page.

@jbalboni , re. that question you had re. removing the label tag altogether, is it still relevant?  I've since made the component require either label or ariaLabelledBy property, not both.

jbalboni:
@tlei123 It is, yeah. We still have the `label` element being output in the render method, even if the label prop is empty. I *think* we should remove the element when that is the case, but I'd want to know Trevor's opinion (and also if it breaks anything about how the checkbox displays).

packages/formation-react/src/components/ErrorableCheckbox/ErrorableCheckbox.jsx
```diff
@@ -51,6 +51,7 @@ class ErrorableCheckbox extends React.Component {
           </span>
         )}
         <input
+          aria-labelledby={this.props.ariaLabelledBy}
```
1Copenut: As long as this `aria-labelledby` attribute string matches the ID being passed into the heading above it, and those IDs are unique, this should be fine. I'm not able to get this running locally to verify it, but looking at the code this seems logical and correct. I'll plan to test in Staging when I am back in the office.

tlei123: @jbalboni , @1Copenut , not hitting a solid brick wall trying to remove the label, but quite a thick hedge...

This is a _custom_ checkbox, and relies on the label for both displaying a “fake” checkbox (with CSS label::before) and more-importantly the click-handling as well.

The User doesn't really click the input element itself, which is hidden (with CSS opacity:0).  The click is actually on the label (empty or not) and relies on its for attribute to change the hidden-input value.

I'm just wondering though whether it's worth a major refactor (both styling and functionality) just to remove the label element for this use-case (which seems an edge one to me).  The styling is not just within formation, but rides on top of external uswds stylesheets, making any tweaks more brittle than I'd like.

What are your thoughts?  Any harm done leaving the empty label?

jbalboni: Ah, I had forgotten that that's how the checkbox was implemented. No, I don't think it's worth trying to refactor that, as long as the checkbox seems work for fine when the `label` has no text content.

tlei123: Thx, @jbalboni !  Leaving empty label then for this.

packages/formation-react/src/components/ErrorableCheckbox/ErrorableCheckbox.jsx
```diff
@@ -88,13 +89,56 @@ ErrorableCheckbox.propTypes = {
    */
   name: PropTypes.string,
   /**
-   * Label for the checkbox
+   * Label [string or object] for the checkbox. Either this or ariaLabelledBy is required.
+   * IF you have a SEPARATE HEADING (outside this component) that “labels” this
+   * checkbox, use ariaLabelledBy instead of this.
+   * Screen-readers read sr-only content, which would be redundant if you had a
```
jbalboni: I would remove these lines about `sr-only` content and the example. If we feel like that's necessary info, I think maybe putting it somewhere in the mdx page for this component would make more sense. It's going to hard to read in the prop types table.

tlei123: Done.

1Copenut:
@tzelei I just tested this one with the following combinations:

* MacOS + Safari + VoiceOver
* Win10 + IE11 + JAWS
* Win10 + Chrome + JAWS
* Win10 + Chrome + NVDA
* Win10 + Firefox + NVDA

All worked as expected. Some of the assistive tech announced the heading as clickable when moving by virtual cursor, but listened for the click events and toggled the checkboxes appropriately. Checkboxes were available to toggle, and in the Forms Controls menus.

PR #163, opened 2019-09-04
Update the CMS feature flags documentation

cvalarida:
## Description
The documentation was out of date. This brings it back up to date.

## Testing done
N/A

## Screenshots
N/A

## Acceptance criteria
- [x] The documentation accurately reflects the current state of the CMS feature flags

## Definition of done

### DIFF
--- a/packages/documentation/src/pages/platform/tools/feature-flags.mdx
+++ b/packages/documentation/src/pages/platform/tools/feature-flags.mdx
@@ -55,7 +55,7 @@ localhost: false
   - Making code that's easy to delete ensures that deprecated code is quickly removed once a feature is released.
 - Making features or applications available in production should done only as needed for testing with users.
 
-## Feature flags for Metalsmith code
+## CMS feature flags
 
 Va.gov creates some pages based on content from the VA's Drupal CMS. The CMS has its own content model, which can sometimes change. When those changes modify the existing structure of the content model, the queries and templates in vets-website that expect a different model may break. And because the CMS and vets-website are separate systems with different deployment processes, we can't push up changes in both systems simultaneously. In order to keep the two systems in sync, we need to be able to turn features on and off in vets-website depending on what environment we're in, and update that feature state whenever a cms deployment happens.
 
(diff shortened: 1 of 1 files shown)


packages/documentation/src/pages/platform/tools/feature-flags.mdx
```diff
@@ -55,7 +55,7 @@ localhost: false
   - Making code that's easy to delete ensures that deprecated code is quickly removed once a feature is released.
 - Making features or applications available in production should done only as needed for testing with users.
 
-## Feature flags for Metalsmith code
+## CMS feature flags
```
jbalboni: If you haven't, you might want to search for links to the old header (something like `#feature-flags-for-metalsmith-code`) on other pages.

packages/documentation/src/pages/platform/tools/feature-flags.mdx
```diff
@@ -64,57 +64,16 @@ We've created some infrastructure to make this a little easier to do.
...
-applyFeatureFlags(module);
-```
-
-This helper looks through all of the enabled feature flags and checks to see if there are files with the same name as the current file, but with a feature flag in the file path before the extension.
-
-So, for example, if your file is `healthCarePage.graphql.js` and there's a feature flag called `feature1`, the helper will look for `healthCarePage.graphql.feature1.js`. If `feature1` is turned on, then it will replace the exported module data with the export from the file with the `feature1` suffix.
+When running a new vets-website build, the script will fetch all CMS feature flags from Drupal. This allows the flags to be controlled from within Drupal.
 
-If you want to make a smaller conditional logic change than creating a whole new file, you can import the featureFlags file and check the flags directly:
+Before writing any code to use a new feature flag, it must first be created from within Drupal for all three environments. The feature flags can be found at https://dev.cms.va.gov/admin/config/system/feature_toggle for dev. Staging and production have similar pages.
```
jbalboni: You could probably make that url a link, as well as the Staging and production text in the next sentence.

packages/documentation/src/pages/platform/tools/feature-flags.mdx
rianfowler: We're calling these "query config flags" now. Open to other suggestions: https://docs.google.com/a/adhocteam.us/document/d/1ESEH7P8BYFZdqMEd9ePzN3L8stuH8E8FwBViR8tZvos/edit?disco=AAAADb64FVE&usp=comment_email_discussion&usp_dm=false&ts=5d69a21c

rianfowler: "feature flags" is reserved for the flipper toggles.

cvalarida:
@ncksllvn We've got some plans in place now to fix most of the issues you mentioned.

## Query flag entanglement
This is a really good point. I definitely should have investigated the ramifications of the change to query Drupal for the flags a lot more before I made it. :grimacing: 

### The build
So there are a couple things here:
1) PRs unrelated to anything CMS can break because of missing feature flags because they've been removed in the CMS
2) Engineers can't build locally because the build requires a cache...which they can't get yet
    - You called this one out separately, so I'll keep this separate for now; see below

After talking with @gunsch, we decided the best path forward was to:
1) Stop the build from breaking on feature branches
    - This should still break on `master` builds, though, so we don't deploy "degraded" content
2) Alert the CMS team when we detect missing query flags

We have a ticket to track this work: https://github.com/department-of-veterans-affairs/va.gov-team/issues/1583

### The cache
This one is tricky. I'm not sure what the answer is yet, but we have a ticket for it: https://github.com/department-of-veterans-affairs/va.gov-team/issues/1607

### Preview server
I don't really know what the story is here. How is this affecting the preview servers?

### Running vets-website without SOCKS access
This _should_ have been possible by running `yarn fetch-drupal-content` to fetch the latest cache from the S3 bucket, but because the query flags affect the cache lookup, this is currently broken. :disappointed:

@micahchiang and I were able to get the localhost build to work without needing any content from Drupal or a cache in [this PR](https://github.com/department-of-veterans-affairs/vets-website/pull/10687), however, which should solve the immediate-term problem of getting engineers up and running before they have SOCKS access.

## The case of the disappearing query flags
This is definitely a struggle. I believe the CMS team is working on a fix for this. In the meantime, ensuring that this doesn't affect feature branches (mentioned above) should tide us over.

## Drupal access
By auto-alerting the CMS team, I think we'll mitigate the need for this. The model I want to set up is one that enables that team to take care of all the normal CMS stuff without assistance from platform engineers.

That said, I don't know how your team factors into the mix. If your team needs access and understanding, I think that can be set up.

## Hard-coded query flags
This was definitely easier. If we can't figure out the caching issue, I think I'm going to advocate for moving back to this. At this point, the benefit doesn't seem worth the cost.

gunsch:
> This should have been possible by running yarn fetch-drupal-content to fetch the latest cache from the S3 bucket, but because the query flags affect the cache lookup, this is currently broken. 😞

One of my proposals yesterday was to have all the query flag values evaluate to "false" for the cache key generation, so that changing a query flag value doesn't affect "last known good content". If we're still OK with that path forward, doesn't that fix this problem as well?

> This should still break on master builds, though, so we don't deploy "degraded" content

Hmm, I thought we were gonna disable the "breaking" entirely so that it doesn't keep halting deployments? Broken deployments primarily hurts our team, whereas just not rendering the content they wanted aligns the consequence with the action (content team wants to content out, they have to get it right).

> Hard-coded query flags

Updating things in two places isn't great, but the realities of this situation aren't great either. I'd be supportive if you decide to make this call.

ncksllvn:
The work with the CMS is already very complex, and to me, the query flags have only added complexity. Thoughts coming to mind right now, although I could be wrong on this any step of the way - 

1.  I don't know the consequences of having the query flags fallback to false if unavailable.
    - Could that lead to outdated content being built? It seems like it could, unless we had master always require latest CMS and not fall back to cache. For instance - master builds off cache -> daily deploy fails -> we can't get the CMS data to build -> we execute a build without the pre_release job for getting latest content -> outdated content makes its way into prod. 
     - Since the CMS cache on S3 relies on an MD5 of the GraphQL GetAllPages query, and the query flags modify that, would the cache still be available if we're falling back to false for all of them?
1. On the front-end, how can we see when we can eliminate the tech debt introduced by a query flag? A hardcoded flag listed under every environment pretty clearly tells us that this flag can be torn down, while a flag from an API suggests that it's still subject to change. 
1. Does this make sense conceptually for a client/API relationship? I know the CMS is more than just any API to the website, but if we're trying to model things after a typical client/API relationship, would we ever have a list of query flags come from an API to tell us how to modify the subsequent API requests?

This is all introduced by the difficulty we've run into mapping vets-website environments to CMS environments. From what we've learned here, I'm going to back to asking the big questions -

1. Why can't the CMS build all environments off of master, like we do with Vets-API?
1. If we can't do that, can we point all vets-website environments to a CMS Prod?

jbalboni:
> 3. Does this make sense conceptually for a client/API relationship? I know the CMS is more than just any API to the website, but if we're trying to model things after a typical client/API relationship, would we ever have a list of query flags come from an API to tell us how to modify the subsequent API requests?

I think the issue is that the Drupal api doesn't follow the typical practices of a GraphQL or REST api, and config flags are a reaction to that. If it were a REST api, you'd expect changes in the returned data to bump an api version in the url. In GraphQL, the expectation is that you add new fields and deprecate old ones (without removing them). The current Drupal process doesn't follow this and so you're sort of stuck reacting to it.

Even if we were mapping to one environment (which would reduce a lot of complexity in general), you'd still have the need for query flags, because the api is not stable over time.

cvalarida:
@gunsch 
Ah! Thanks for the clarification.

## Evaluating false for cache key generation
**TL;DR:** I think this would work.

I just took entirely too long figuring this all out again.

Caching this way, the build process would look something like this:
1) We query for feature flags and get a result `{ foo: true, bar: false }`
2) We construct the query with these flags
3) Using this query, we fetch the content
4) At the end of the Jenkins pipeline, we cache the result
    - The cache key would be generated by constructing the query where all the flags are false 
    - Since the flags and query result are cached with each other, this is how we find the right cache irrespective of the feature flags, which are prone to flapping
    - :thinking: If the liquid templates are out-of-sync with the cached data, will this affect anything?
      - If the templates use flags that don't exist, they'll just evaluate to falsey and that should fail safely
      - If the query result has extra fields (query flags that were true) that the template doesn't use...that should be fine?
      - Am I missing any edge cases?

I think you're right. This should solve the problem we're running into with the `fetch-drupal-content` script.

## Disable missing query flags from breaking anything
Ahh, right right. That does ring a bell now. I forgot that's what we landed on. I was conflating this with broken links, which we'll let through the feature branch builds, but not the `master` builds.

---

@ncksllvn 
...I'll have to get back to you on Monday. I've got to go now. :grimacing: 

I think the short story is that yes, we'll build outdated content, but it _should_ be fine—like falling back to any other cache. I'm a little nervous about how the liquid templates play into that, but I think that's it. I'm even debating caching the templates alongside the query results. :sweat_smile:

PR #167, opened 2019-09-11
Update internal-tools.mdx

ricetj:
## Description
Fixing a change I made so it does not so bad


ricetj:
@jefflbrauer I just made a small change so it doesn't look so broken. I don't want to change or throw off your flow, so that is why I am tagging you.

jefflbrauer:
Did you mean to tag Jeff Balboni or Barnes?

ricetj:
> Did you mean to tag Jeff Balboni or Barnes?

yeah sorry about that wrong tag.

PR #170, opened 2019-09-23
Fully enable shift-tabbing in Modal components

erikphansen:
## Description
Previously shift-tab only worked until focus reached the first
focusable element in the modal. Hitting shift-tab when focus was on the
first element in the modal resulted in effectively a no-op as focus was
programmatically set to the first focusable element. With this change a
shift-tab while the first element on the modal has focus will move focus
to the _last_ element in the modal, as you would expect.

## Testing done
Local in vets-website

## Screenshots
**GIF showing ability to tab forwards and backwards within the modal correctly**
modal-shift-tab

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/formation-react/package.json
+++ b/packages/formation-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "4.10.0",
+  "version": "4.10.1",
   "description": "React implementation of patterns in Formation",
   "keywords": [
     "react",
--- a/packages/formation-react/src/components/modal/Modal.jsx
+++ b/packages/formation-react/src/components/modal/Modal.jsx
@@ -3,11 +3,12 @@ import React from 'react';
 import classNames from 'classnames';
 
 const ESCAPE_KEY = 27;
+const TAB_KEY = 9;
 
 class Modal extends React.Component {
   constructor(props) {
     super(props);
-    this.state = { lastFocus: null };
+    this.state = { lastFocus: null, isTabbingBackwards: false };
   }
 
   componentDidMount() {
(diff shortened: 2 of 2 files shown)


packages/formation-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "4.10.0",
+  "version": "4.10.1",
```
erikphansen: I'm calling this a bug-fix rather than new feature.

packages/formation-react/src/components/modal/Modal.jsx
```diff
@@ -29,9 +30,9 @@ class Modal extends React.Component {
   }
 
   setupModal() {
-    this.applyFocusToModal();
+    this.applyFocusToFirstModalElement();
```
erikphansen: I renamed this existing method since we will now be programmatically focusing the last element in the modal as well as the first.

packages/formation-react/src/components/modal/Modal.jsx
```diff
@@ -29,9 +30,9 @@ class Modal extends React.Component {
...
     document.body.classList.add('modal-open');
-    document.addEventListener('keyup', this.handleDocumentKeyUp, false);
+    document.addEventListener('keydown', this.handleDocumentKeyDown, false);
```
erikphansen: switching to a `keydown` here. `keydown` is triggered _before_ `focus` while `keyup` is triggered _after_ `focus` and we want to take some action before the browser has focused on the next element.

erikphansen:
@kelsonic I mentioned I'd tag you on this small PR to the Formation React components.

PR #173, opened 2019-09-26
Generated pages in documentation site with va.gov-team repo as source

U-DON:
## Description
This is a demonstration of populating the documentation site by pulling documents from the va.gov-team repo. In this example, I only pulled the directory `platform/working-with-vsp`, but we can choose to pull from additional paths or a higher level directory.

The pages from GitHub are currently indexed for search only by their titles, which is based on the very first heading in the document. As an initial proposal for indexing the pages with tags to improve searchability, we could possibly add frontmatter to the source GitHub documents like so:

```
---
tags: review, pr, merge
---

**ACTUAL CONTENT**
```

The template for the generated pages currently just incorporates the sidebar layout that's used across the site. We can later refactor to have the sidebar content reflect the repo's file hierarchy. Furthermore, the bottom of each of these generated pages will include a link to the source document.

## Testing done
Local testing. Functionality can be verified by searching by the first heading of any of the pages under the `platform/working-with-vsp` directory.

For example, to look up this [doc on deployment policy](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/working-with-vsp/policies-work-norms/deployment-policy.md), search any term in "Automated Deploys", which is the document's first header.


U-DON:
This should be accessible on the site, whether you use the search bar to find the page by its title or directly access the URL. In my example for Testing Done, you would find the page by searcing "Automated Deploy" or going directly to the URL, `/platform/working-with-vsp/policies-work-norms/deployment-policy`. Note that I set its path in the Gatsby to be the path of the document in the original repo but with the file extension trimmed.

I don't know off the top of my head how it would be added to the sidebar, but I imagine it'll involve some GraphQL query wrangling to gather all the nodes that were created from pulling from the va.gov-team repo.

PR #178, opened 2019-10-09
AnyCorp - Adding a React Modal view

1Copenut:
## Description
Several teams have asked for a modal dialog to be added to the AnyCorp learning tool to illustrate requirements to use this UI pattern accessibly. I used the [React Modal](https://github.com/reactjs/react-modal) because it is well supported and captures all of the accessibility features I want.

## Testing done
* Manual testing with VoiceOver + Safari, JAWS + Chrome, NVDA + Firefox
* Keyboard testing to determine focus management, focus return on close, and proper close button

## Definition of done

---
Screen Shot 2019-10-09 at 3 06 08 PM

### DIFF
--- a/packages/accessible-react-app/package.json
+++ b/packages/accessible-react-app/package.json
@@ -12,6 +12,7 @@
     "start": "yarn webpackDev",
     "server": "node server.js",
     "webpackDev": "webpack-dev-server --open --mode development --config webpack.dev.config --hot",
+    "webpackDev:lan": "webpack-dev-server --open --mode development --host 0.0.0.0 --config webpack.dev.config --hot",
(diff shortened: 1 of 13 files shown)


1Copenut:
@rianfowler this is a great point. Given it's such a small adjustment, I'm going to refactor a bit before merging, and just have the modal be information, not a form.

erikphansen:
but given the fact that we do have forms in modals on va.gov, is it worth showing an example of doing that well (even though Rian hates it)?

1Copenut:
> but given the fact that we do have forms in modals on va.gov, is it worth showing an example of doing that well (even though Rian hates it)?

And a very compelling counter-argument. I'll keep it for now, and use it as a talking point for what we should try not to do, in design *and* development.

PR #179, opened 2019-10-09
[Additional Info] Implement expand/collapse events

ncksllvn:
## Description
This PR changes the events triggered by the Additional Info component to use `expand` and `collapse` events instead of a single `button-clicked`, and triggers the event from the additional-info container element instead of the generated button. This way, the `event.target` is a reference to the original DOM element, so consumers can access the container's dataset through `event.target.dataset` and this component no longer needs to explicitly pass the `detail` property in the event. I separated the `button-clicked` event into `expand`/`collapse` because it seemed like the cleanest way for a listener to determine what happened, as opposed to checking the `aria-expanded` property or by passing in a new property in the `detail` like `expanded: true/false`.

This work comes from standardizing our design system's events. Details for this work are here, https://github.com/department-of-veterans-affairs/va.gov-team/issues/1541.

---
Companion vets-website PR -https://github.com/department-of-veterans-affairs/vets-website/pull/10876

Example usage:
```js
  document.body.addEventListener(
    '@department-of-veterans-affairs/formation/additional-info/expand',
    () => recordEvent({ event: 'nav-additionalInfo-expand' }),
  );
```

## Testing done
Locally testing using GTM browser extension

## Screenshots
image

## Acceptance criteria
- [ ] Events are standardized

## Definition of done


packages/formation/js/additional-info.js
```diff
@@ -43,22 +49,28 @@ export default function createAdditionalInfoWidget() {
...
+          isExpanded,
+        );
+        chevron.classList.toggle('open', isExpanded);
 
-        button.dispatchEvent(additionalInfoClickEvent);
+        if (isExpanded) {
+          el.dispatchEvent(additionalInfoExpandEvent);
+        } else {
+          el.dispatchEvent(additionalInfoCollapseEvent);
+        }
```
ncksllvn: Really open to discussion here. Is separating the event discouraged? Should the button be the trigger, instead of the container, since that's where the event originated?

packages/formation/js/additional-info.js
```diff
@@ -43,22 +49,28 @@ export default function createAdditionalInfoWidget() {
...
       const button = qS(el, 'button');
 
-      const additionalInfoClickEvent = document.createEvent('Event');
-      additionalInfoClickEvent.initEvent(
-        '@department-of-veterans-affairs/formation/additional-info/button-clicked',
-        true,
-        true,
-      );
-      additionalInfoClickEvent.detail = { titleText, dataset };
+      const additionalInfoExpandEvent = document.createEvent('Event');
```
kelsonic: If I'm not mistaken, I think MDN prefers devs using `new Event()` over `createEvent` as some browsers may in the near-future deprecate those methods. https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent

packages/formation/js/additional-info.js
kelsonic: Exactly like you said, I would do `button.dispatchEvent(additionalInfoExpandEvent);`since that's where the event is propagating from :) Separating the event with the conditional makes sense to me though!

packages/formation/js/additional-info.js
ncksllvn: Unfortunately, `new Event` isn't supported by IE - https://developer.mozilla.org/en-US/docs/Web/API/Event/Event - so we're stuck with this method 😞

kelsonic: Oh wow, such conflicting info from MDN then... 😂Thanks IE! ✊

PR #190, opened 2019-11-26
Refactor a bit and change onSort prop to onHeaderClick for accuracy

kelsonic:
## Description
This PR includes some backwards-compatible refactoring for the `<SortableTable />` component. I

It also introduces a deprecation warning when the prop called `onSort` is used, which is called whenever a header is clicked (it doesn't actually do any sorting). This prop is **not currently used in [vets-website](https://github.com/department-of-veterans-affairs/vets-website).

It also introduces a new prop called `onHeaderClick`, which is a more accurate name for the `onSort` prop and the intent is that this prop will replace `onSort` long-term.

## Testing done
Unit tests.

## Screenshots
N/A

## Acceptance criteria
- [x] Add deprecation warning for the `onSort` prop.
- [x] Add `onHeaderClick` prop to replace `onSort` eventually.

## Definition of done

### DIFF
--- a/packages/formation-react/src/components/SortableTable/SortableTable.jsx
+++ b/packages/formation-react/src/components/SortableTable/SortableTable.jsx
@@ -58,7 +58,7 @@ class SortableTable extends Component {
     this.props.onHeaderClick(value, order);
   }
 
-  makeHeader = (field) => {
+  renderHeader = (field) => {
     if (field.nonSortable) {
       return <th key={field.value}>{field.label}</th>;
     }
@@ -97,7 +97,7 @@ class SortableTable extends Component {
     );
   }
 
-  makeRow(item) {
+  renderRow = (item) => {
     return (
       <tr key={item.id} className={item.rowClass}>
         {this.props.fields.map(field => (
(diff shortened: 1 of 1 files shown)


kelsonic:
The props need to be updated here in the docs, but not sure how to do that: https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/visual-design/components/sortabletable/ Do you know where/how I should update that by any chance? @ncksllvn

packages/formation-react/src/components/SortableTable/SortableTable.jsx
```diff
@@ -1,19 +1,64 @@
...
+    // This is a legacy prop that is being renamed to `onHeaderClick`.
+    if (onSort) {
+      console.warn('WARNING: The `onSort` prop is being deprecated in favor of the `onHeaderClick` prop for <SortableTable />.');
+      onSort(value, order);
+    }
 
-  handleSort(value, order) {
-    return () => this.props.onSort(value, order);
+    // This replaces `this.props.onSort`.
+    this.props.onHeaderClick(value, order);
```
ncksllvn: I could be reading it wrong, but do we need an `else` here, so that we don't call both `onSort` and `onHeaderClick`?

kelsonic: So I actually thought about returning early on [this line](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/190/files/b674f7b5e5c85017a97495dae15f7d9e43611d38#diff-1ce346f7eeb87e7b181f44b2dabb5566R54) but actually decided against it since there's no errors if they decide to use both in theory! :)

kelsonic:
Closed this PR as it is now included in [this PR](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/191).

PR #191, opened 2019-11-26
Kelsonic 3725 update sortabletable

kelsonic:
## Description
**WARNING: THIS PR INCLUDES [THIS PR](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/190).**

**Original Issue:** https://github.com/department-of-veterans-affairs/va.gov-team/issues/3701

This PR adds a new prop called `renderCustomCell` for the `<SortableTable />` component, which is a function that receives the row item as an argument and returns back custom JSX to render within the table cell.

## Testing done
Unit tests.

## Screenshots
N/A

## Acceptance criteria
- [x] Add a new `renderCustomCell` prop.

## Definition of done

### DIFF
--- a/packages/formation-react/src/components/SortableTable/SortableTable.jsx
+++ b/packages/formation-react/src/components/SortableTable/SortableTable.jsx
@@ -58,7 +58,7 @@ class SortableTable extends Component {
     this.props.onHeaderClick(value, order);
   }
 
-  makeHeader = (field) => {
+  renderHeader = (field) => {
     if (field.nonSortable) {
       return <th key={field.value}>{field.label}</th>;
     }
@@ -97,7 +97,7 @@ class SortableTable extends Component {
     );
   }
 
-  makeRow(item) {
+  renderRow = (item) => {
     return (
       <tr key={item.id} className={item.rowClass}>
         {this.props.fields.map(field => (
@@ -108,8 +108,8 @@ class SortableTable extends Component {
   }
 
   render() {
-    const headers = this.props.fields.map(this.makeHeader);
-    const rows = this.props.data.map(this.makeRow);
+    const headers = this.props.fields.map(this.renderHeader);
+    const rows = this.props.data.map(this.renderRow);
     const tableClass = classNames('va-sortable-table', this.props.className);
 
     return (
--- a/packages/formation-react/src/components/SortableTable/SortableTable.jsx
+++ b/packages/formation-react/src/components/SortableTable/SortableTable.jsx
@@ -1,19 +1,69 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { Component } from 'react';
 import classNames from 'classnames';
 
-class SortableTable extends React.Component {
-  constructor(props) {
-    super(props);
-    this.makeHeader = this.makeHeader.bind(this);
-    this.makeRow = this.makeRow.bind(this);
+class SortableTable extends Component {
+  static propTypes = {
+    className: PropTypes.string,
+
+    /**
+     * Field value to sort by in either ascending or descending order.
+     * The `value` must be one of the values in the `fields` prop.
+     */
+    currentSort: PropTypes.shape({
+      value: PropTypes.string.isRequired,
+      order: PropTypes.oneOf(['ASC', 'DESC']),
+    }).isRequired,
+
+    /**
+     * Mappings of header labels to properties on the objects in `data`.
+     */
+    fields: PropTypes.arrayOf(
+      PropTypes.shape({
+        label: PropTypes.string.isRequired,
+        value: PropTypes.string.isRequired,
+      }),
+    ).isRequired,
+    /**
+     * Each object represents data for a row.
+     * An optional class may be provided to style specific rows.
+     */
+    data: PropTypes.arrayOf(
+      PropTypes.shape({
+        id: PropTypes.oneOfType([PropTypes.number, PropTypes.string]).isRequired,
+        values: PropTypes.arrayOf(
+          PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+        ),
+        rowClass: PropTypes.string,
+      }),
+    ).isRequired,
+
+    // DEPRECATING: A callback for when a header is clicked.
+    onSort: PropTypes.func,
+    // A callback for when a header is clicked.
+    onHeaderClick: PropTypes.func,
+
+    /**
+     * Each object represents data for a row.
+     * An optional class may be provided to style specific rows.
+     */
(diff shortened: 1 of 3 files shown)


ncksllvn:
Do we need `renderCustomCell` specifically for the paper icon under the `VA form number` column?

packages/formation-react/src/components/SortableTable/SortableTable.jsx
```diff
@@ -1,19 +1,65 @@
...
+    // This is a legacy prop that is being renamed to `onHeaderClick`.
+    if (onSort) {
+      console.warn('WARNING: The `onSort` prop is being deprecated in favor of the `onHeaderClick` prop for <SortableTable />.');
+      onSort(value, order);
+    }
 
-  handleSort(value, order) {
-    return () => this.props.onSort(value, order);
+    // This replaces `this.props.onSort`.
+    this.props.onHeaderClick(value, order);
```
rianfowler: what happens if `onHeaderClick` is not defined?

packages/formation-react/src/components/SortableTable/SortableTable.jsx
```diff
@@ -52,21 +98,39 @@ class SortableTable extends React.Component {
     );
   }
 
-  makeRow(item) {
-    const cells = this.props.fields.map(field => (
-      <td key={`${item.id}-${field.value}`}>{item[field.value]}</td>
-    ));
+  renderCell = (item, field) => {
+    const renderCustomCell = get(item, 'renderCustomCell');
```
rianfowler: is `get` defined in this context?

packages/formation-react/src/components/SortableTable/SortableTable.jsx
```diff
@@ -52,21 +98,39 @@ class SortableTable extends React.Component {
...
+
+    // If the item has a `renderCustomCell` property, execute it.
+    if (renderCustomCell) {
+      return renderCustomCell(field);
```
rianfowler: `return renderCustomCell ? renderCustomCell(field) : item[field.value];`

packages/formation-react/src/components/SortableTable/SortableTable.jsx
kelsonic: Great catch!

packages/formation-react/src/components/SortableTable/SortableTable.jsx
kelsonic: Another great catch!

packages/formation-react/src/components/SortableTable/SortableTable.jsx
kelsonic: I actually wrote it with the `if` statement so that it'll be easier to add on conditionals in the future if we want while still making the code readable :)

kelsonic:
I actually tried to test this with `yalc` but it looks like the only place this component is used currently is for the new find VA forms page!

It looks like it's also used here: `vets-website/src/applications/personalization/dashboard/containers/MessagingWidget.jsx:75` but that actually *doesn't render since it's commented out* here: `src/applications/personalization/dashboard/components/ManageYourVAHealthCare.jsx:96`

kelsonic:
@ncksllvn 

> Can we add a unit test for the renderCell and possibly onHeaderClick?

After further discussion and investigation, it seemed like `renderCustomCell` would actually make things more complex than just rendering JSX with the table items' value. For now there doesn't seem to be performance issues with this approach as we are not currently using stateful components within table cells.

> We should update the SortableTable.mdx to include the onHeaderClick prop

Done! ✅ 

> We need to increment the package.json version to publish this to NPM. We can increment the minor version as-is, or remove onSort altogether and do a major version bump. The latter seems like a lot but we're removing support so it makes sense.

Done! ✅As `onSort` does not seem to be used at all, I removed support for it altogether and did a major version bump.

packages/formation-react/src/components/SortableTable/SortableTable.jsx
```diff
@@ -1,19 +1,62 @@
...
+    onHeaderClick: PropTypes.func,
+  };
+
+  onHeaderClick = (value, order) => () => {
+    const { onHeaderClick } = this.props;
+
+    // This is a legacy prop that is being renamed to `onHeaderClick`.
+    if (onSort) {
+      console.error('WARNING: The `onSort` prop is now deprecated in favor of the `onHeaderClick` prop for <SortableTable />.');
+    }
```
ncksllvn: Since we're bumping the major version, we can remove support for `onSort`.

packages/formation-react/src/components/SortableTable/SortableTable.mdx
```diff
@@ -37,27 +37,28 @@ import SortableTable from '@department-of-veterans-affairs/formation-react/Sorta
...
-        name: 'Bob',
+        name: <span>Bob</span>,
         email: 'bob@bob.com',
         message:'Hola',
         rowClass: 'class'
       },
       {
         id: 3,
-        name: 'Tom',
+        name: <span>Tom</span>,
```
ncksllvn: This is just showing that JSX is optional, correct?

kelsonic: Correct! :)

packages/formation-react/src/components/SortableTable/SortableTable.jsx
kelsonic: All this does is in case an `onSort` prop is provided, they will get an error noting that it's now deprecated so it doesn't just silently fail on them :)

ncksllvn: I think it's okay for it to fail silently since we're bumping the major version. React will warn them about an unrecognized prop anyway.

kelsonic: I don't think PropTypes catches props that aren't included (unrecognized props), but will change this so it fails silently! 👍

packages/formation-react/src/components/SortableTable/SortableTable.jsx
```diff
@@ -1,19 +1,57 @@
...
+    // A callback for when a header is clicked.
+    onHeaderClick: PropTypes.func,
+  };
+
+  onHeaderClick = (value, order) => () => {
+    const { onHeaderClick } = this.props;
+
+    // This replaces `this.props.onSort`.
+    if (onHeaderClick) {
+      onHeaderClick(value, order);
```
ncksllvn: You could also make a default prop for `onHeaderClick` as an empty function.

PR #201, opened 2020-01-03
Remove setState from unmount

bradl3yC:
## Description
This will fix a bug in veterans website caused by modal component
https://github.com/department-of-veterans-affairs/va.gov-team/issues/2427

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/formation-react/package.json
+++ b/packages/formation-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "5.0.0",
+  "version": "5.0.1",
   "description": "React implementation of patterns in Formation",
   "keywords": [
     "react",
--- a/packages/formation-react/src/components/modal/Modal.jsx
+++ b/packages/formation-react/src/components/modal/Modal.jsx
@@ -42,7 +42,6 @@ class Modal extends React.Component {
   teardownModal() {
     if (this.state.lastFocus) {
       this.state.lastFocus.focus();
-      this.setState({ lastFocus: null });
     }
     document.body.classList.remove('modal-open');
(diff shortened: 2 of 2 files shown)


packages/formation-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "5.0.0",
+  "version": "5.0.1",
```
ncksllvn: I think you need to do a `yarn install` to update the `yarn.lock`

packages/formation-react/src/components/modal/Modal.jsx
```diff
@@ -42,7 +42,6 @@ class Modal extends React.Component {
   teardownModal() {
     if (this.state.lastFocus) {
       this.state.lastFocus.focus();
-      this.setState({ lastFocus: null });
```
kelsonic: Are there any drawbacks to not setting `lastFocus` to `null` from this line: https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/201/files#diff-7018d1e165f9f39d7cddc546e8331f62R22

bradl3yC:
@ncksllvn I do not have access to the NPM.  Are you able to publish this?

packages/formation-react/src/components/modal/Modal.jsx
bradl3yC: @kelsonic No I don't think so because this only gets called if the next props is not visible.  That should only happen as the modal is exited and any props passed in come from vets website / redux.  If the modal is opened again, it should hit the constructor and be set to null on mount.

PR #202, opened 2020-01-06
bradl3yC 4560 Reset state and wait for teardown complete

bradl3yC:
## Description
- [x] - Add a pause before updating component to allow for modal teardown func to complete
- [x] - Reset state back to null before closing

## Testing done
- [x] - Tested locally using yalc - performs as needed

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/formation-react/src/components/modal/Modal.jsx
+++ b/packages/formation-react/src/components/modal/Modal.jsx
@@ -26,6 +26,7 @@ class Modal extends React.Component {
   componentWillUnmount() {
     if (this.props.visible) {
       this.teardownModal();
+      this.setState({ lastFocus: null })
     }
   }
 
@@ -45,7 +46,6 @@ class Modal extends React.Component {
       setTimeout(() => {
         this.state.lastFocus.focus();
       }, 0)
-      this.setState({ lastFocus: null })
     }
     document.body.classList.remove('modal-open');
     document.removeEventListener('keydown', this.handleDocumentKeyDown, false);
--- a/packages/formation-react/src/components/modal/Modal.jsx
+++ b/packages/formation-react/src/components/modal/Modal.jsx
@@ -17,7 +17,6 @@ class Modal extends React.Component {
 
   componentDidUpdate(prevProps) {
     if (!prevProps.visible && this.props.visible) {
-      this.setState({ lastFocus: document.activeElement });
       this.setupModal();
     } else if (prevProps.visible && !this.props.visible) {
       this.teardownModal();
@@ -31,6 +30,7 @@ class Modal extends React.Component {
   }
 
   setupModal() {
+    this.setState({ lastFocus: document.activeElement });
     this.applyFocusToFirstModalElement();
     document.body.classList.add('modal-open');
     document.addEventListener('keydown', this.handleDocumentKeyDown, false);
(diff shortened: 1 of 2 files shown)


packages/formation-react/src/components/modal/Modal.jsx
```diff
@@ -25,7 +25,9 @@ class Modal extends React.Component {
 
   componentWillUnmount() {
     if (this.props.visible) {
-      this.teardownModal();
+      setTimeout(() => {
+        this.teardownModal()
+      }, 0);
     }
```
bradl3yC: This will hold up closing the modal / halting the teardown until after teardown is complete.

packages/formation-react/src/components/modal/Modal.jsx
```diff
@@ -41,14 +41,17 @@ class Modal extends React.Component {
 
   teardownModal() {
     if (this.state.lastFocus) {
-      this.state.lastFocus.focus();
+      setTimeout(() => {
```
erikphansen: this seems like a good candidate for a comment explaining why we need this event-loop hack

bradl3yC: @erikphansen added

packages/formation-react/src/components/modal/Modal.jsx
```diff
@@ -41,14 +41,18 @@ class Modal extends React.Component {
...
+        this.state.lastFocus.focus();
+      }, 0)
     }
     document.body.classList.remove('modal-open');
     document.removeEventListener('keydown', this.handleDocumentKeyDown, false);
     document.removeEventListener('focus', this.handleDocumentFocus, true);
     if (this.props.clickToClose) {
       document.removeEventListener('click', this.handleDocumentClicked, true);
     }
+    this.setState({ lastFocus: null })
```
erikphansen: this line needs to be moved into the timeout callback as well or it creates a bug with management of the buttons that launch the DD modal

erikphansen: but that seems to lead to a react error indicating a memory leak. looking into a better fix.

bradl3yC: @erikphansen fixed

packages/formation-react/src/components/modal/Modal.jsx
```diff
@@ -41,7 +41,11 @@ class Modal extends React.Component {
 
   teardownModal() {
     if (this.state.lastFocus) {
-      this.state.lastFocus.focus();
+      // Ensure last focus is set before completing modal teardown
+      setTimeout(() => {
+        this.state.lastFocus.focus();
+      }, 0)
+      this.setState({ lastFocus: null })
```
erikphansen: this still leads to a problem with the buttons that launch the DD modal :(

`Uncaught TypeError: Cannot read property 'focus' of null`

packages/formation-react/src/components/modal/Modal.jsx
bradl3yC: @erikphansen just pushed a change that seems to solve for all.. we only clear state on the last step of the component unmounting

packages/formation-react/src/components/modal/Modal.jsx
```diff
@@ -93,9 +97,6 @@ class Modal extends React.Component {
     if (focusableElement) {
       // we only want to set `lastFocus` when the modal first pops up, not every
```
erikphansen: I guess we can kill this comment as well

PR #207, opened 2020-01-23
Add/modify props for <Breadcrumbs /> and <Pagination />

kelsonic:
## Description
This PR deprecates the prop `customClasses` for the `<Breadcrumbs />` component. It also adds support for a new prop called `className` to replace `customClasses`. It also updates its documentation with the new prop.

[This is the only place in `vets-website` where this prop is currently used](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/vaos/components/Breadcrumbs.jsx#L7). I will send a follow-up PR with the version bump to update this loc 🙂 

#### PR Additions!

This PR now also adds support for a `className` prop for the `<Pagination />` component. It also updates its documentation with the new prop.

## Screenshots
image (6)

## Acceptance criteria
- [x] Update `customClasses` prop to `className` for `<Breadcrumbs />`.
- [x] Add `className` prop for `<Pagination />`.

## Definition of done

### DIFF
--- a/packages/formation-react/src/components/Breadcrumbs/Breadcrumbs.jsx
+++ b/packages/formation-react/src/components/Breadcrumbs/Breadcrumbs.jsx
@@ -28,16 +28,11 @@ class Breadcrumbs extends Component {
   };
 
   render() {
-    const { ariaLabel, customClasses, className, id, listId, mobileFirstProp } = this.props;
+    const { ariaLabel, className, id, listId, mobileFirstProp } = this.props;
 
     // Derive IDs.
     const breadcrumbId = id || uniqueId('va-breadcrumbs-');
     const breadcrumbListId = listId || uniqueId('va-breadcrumbs-list-');
-
-    // Warn if they are using legacy props.
-    if (customClasses) {
-      console.warn('The prop `customClasses` for the <Breadcrumbs /> component has been deprecated. Please use the prop `className` instead.')
-    }
 
     return (
       <nav
--- a/packages/formation-react/src/components/Breadcrumbs/Breadcrumbs.jsx
+++ b/packages/formation-react/src/components/Breadcrumbs/Breadcrumbs.jsx
@@ -1,6 +1,7 @@
+// Node modules.
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import React from 'react';
-import classNames from 'classnames';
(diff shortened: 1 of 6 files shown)


packages/formation-react/src/components/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -39,16 +28,26 @@ class Breadcrumbs extends React.Component {
...
+    const { ariaLabel, customClasses, className, id, listId, mobileFirstProp } = this.props;
+
+    // Derive IDs.
+    const breadcrumbId = id || uniqueId('va-breadcrumbs-');
+    const breadcrumbListId = listId || uniqueId('va-breadcrumbs-list-');
+
+    // Warn if they are using legacy props.
+    if (customClasses) {
+      console.warn('The prop `customClasses` for the <Breadcrumbs /> component has been deprecated. Please use the prop `className` instead.')
+    }
```
ncksllvn: I think a warning would be nice if it were a heads-up about removing support in an upcoming release, but since we're bumping the minor version I don't think this is really necessary. Up to you though if you feel strongly.

ncksllvn:
Oh wait, @kelsonic I think there _might_ be an `.mdx` doc that needs updating for the doc site.

packages/formation-react/src/components/Breadcrumbs/Breadcrumbs.jsx
kelsonic: That makes sense! There was just [this one place in `vets-website`](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/vaos/components/Breadcrumbs.jsx#L7) where that prop is still used, so as long as the PR to bump formation-react on vets-website also includes a fix for that we don't need the warning imo 🙂

packages/formation-react/src/components/Pagination/Pagination.jsx
```diff
@@ -140,7 +150,10 @@ class Pagination extends React.Component {
     });
 
     return (
-      <div className="va-pagination">
+      <div className={classNames({
+        'va-pagination': true,
+        [className]: className,
```
kelsonic: @ncksllvn is there a cleaner way to write this?

PR #235, opened 2020-04-24
Add scrollToAlert to componentDidMount in AlertBox

sanlouise:
## Description
When passed both `showOnScroll` and `isVisible`, we need the Alert box to come into view. When statically passing either of those props to the component, we need to run the scroll in `componentDidMount` rather than in `componentDidUpdate`.

## Testing done
The solution works locally.

## Acceptance criteria
- [x] Scroll to the AlertBox when it appears in the DOM

## Definition of done

### DIFF
--- a/packages/formation-react/src/components/AlertBox/AlertBox.jsx
+++ b/packages/formation-react/src/components/AlertBox/AlertBox.jsx
@@ -3,9 +3,8 @@ import React from 'react';
 import classNames from 'classnames';
 
 class AlertBox extends React.Component {
-  constructor(props) {
-    super(props);
-    this.scrollToAlert = this.scrollToAlert.bind(this);
+  componentDidMount() {
+    this.scrollToAlert();
   }
 
   shouldComponentUpdate(nextProps) {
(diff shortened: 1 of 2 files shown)


packages/formation-react/src/components/AlertBox/AlertBox.jsx
```diff
@@ -16,21 +15,23 @@ class AlertBox extends React.Component {
...
+    }
+
     if (this._ref && !isInView) {
       this._ref.scrollIntoView({
         block: 'end',
         behavior: 'smooth',
       });
     }
-  }
+  };
```
kelsonic, line 34: Did the linter add this `;`? 🙂

packages/formation-react/src/components/AlertBox/AlertBox.jsx
```diff
@@ -16,21 +15,23 @@ class AlertBox extends React.Component {
...
-      this.scrollToAlert();
-    }
+    this.scrollToAlert();
   }
 
-  scrollToAlert() {
+  scrollToAlert = () => {
     const isInView = window.scrollY <= this._ref.offsetTop;
 
+    if (!this.props.isVisible || !this.props.scrollOnShow) {
```
kelsonic, line 24: I see, so we escape early in `scrollToAlert` so we don't need to have the conditional logic in `componentDidMount` and `componentDidUpdate`, smart! 😎

packages/formation-react/src/components/AlertBox/AlertBox.jsx
erikphansen: it's because it switched from a function declaration to expression I think

PR #245, opened 2020-04-28
Kristinagore 8502

kristinagore:
https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/8502

Do you think these changes answer Jesse's questions? Also, does it make sense to include where FEATURE_FLAG_NAMES is imported from in question 8? Thanks!

### DIFF
--- a/packages/documentation/src/pages/platform/tools/feature-toggles.mdx
+++ b/packages/documentation/src/pages/platform/tools/feature-toggles.mdx
@@ -56,7 +56,7 @@ Follow these steps to add and use a new feature toggle in vets-website:
 
 5. Navigate to [http://localhost:3000/flipper/features](http://localhost:3000/flipper/features) and verify that you see your new feature name. If not, restart your rails server.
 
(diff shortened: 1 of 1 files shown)


packages/documentation/src/pages/platform/tools/feature-toggles.mdx
```diff
@@ -66,9 +66,9 @@ const FEATURE_FLAG_NAMES = Object.freeze({
 
 7. Submit a PR for each feature. Crosslinking the PRs in a comment will make it easier for the reviewers to check.
 
-8. Use the selector helper to build a selector for your feature toggle.
+8. Use the selector helper to build a [selector](https://daveceddia.com/redux-selectors/) for your feature toggle. See an example [here](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/gi/containers/ProfilePage.jsx#L118).
```
cvalarida, line 69: We should probably pin the link to a specific commit so the line number doesn't get out-of-sync with what we're trying to point at.
```suggestion
8. Use the selector helper to build a [selector](https://daveceddia.com/redux-selectors/) for your feature toggle. See an example [here](https://github.com/department-of-veterans-affairs/vets-website/blob/f6178ac59b1d691e0f31fb36903b8ba54091b517/src/applications/gi/containers/ProfilePage.jsx#L118).
```

PR #268, opened 2020-05-28
Add initialFocusSelector prop to Modal 

mjlumetta:
## Description

This PR is a small change to allow consumers of the `Modal` component to customize where programmatic focus starts when the modal is first opened. The primary use case of this feature - or at least, the use case I wanted to solve - is to initiate keyboard focus on the modal content rather than the close button without removing the close button. This is a minor improvement for keyboard users. 

Feedback on how useful this change is would be welcome - it's certainly a corner case but seems like an improvement, in my opinion.

## Testing done

* Manually tested with the [crisis line modal](https://github.com/department-of-veterans-affairs/developer-portal/blob/master/src/components/crisisLine/VeteransCrisisLine.tsx) in the Lighthouse developer portal
* Added unit tests for programmatic focus

## Screenshots

Focus on modal open without this change:

ModalWithoutCustomInitialFocus170

Focus on modal open with the change:

ModalWithCustomInitialFocus170

## Acceptance criteria
- [X] If the `initialFocusSelector` prop is provided and matches an element other than the first element that matches the `focusSelector` prop (which is the close button, with the default props), focus when the modal opens is on the first element matching `initialFocusSelector`.

## Definition of done
- [X] Changes have been tested in ~~vets-website~~ developer-portal
  - do the props on the documentation page get generated from `propTypes`, or do they need to be updated somewhere else?
  - original issue (if you have access to Lighthouse on the VA MAX.gov JIRA): https://vajira.max.gov/browse/API-170
- [X] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

### DIFF
--- a/packages/formation-react/package.json
+++ b/packages/formation-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "5.3.0",
+  "version": "5.4.0",
   "description": "React implementation of patterns in Formation",
   "keywords": [
     "react",
--- a/packages/formation-react/src/components/modal/Modal.jsx
+++ b/packages/formation-react/src/components/modal/Modal.jsx
@@ -31,7 +31,7 @@ class Modal extends React.Component {
 
   setupModal() {
     this.setState({ lastFocus: document.activeElement });
-    this.applyFocusToFirstModalElement();
+    this.setInitialModalFocus();
     // NOTE: With this PR (https://github.com/department-of-veterans-affairs/vets-website/pull/11712)
     // we rely on the existence of `body.modal-open` to determine if a modal is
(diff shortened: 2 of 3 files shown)


packages/formation-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "5.2.6",
+  "version": "5.2.7",
```
1Copenut, line 3: I'm always curious about versioning, and teammates' opinions how to bump enhancements. In this case, I feel it might be more of a minor than a fix because of the updated focus method and new prop. Not a huge deal, but maybe worth calling this version 5.3.0.

PR #358, opened 2020-07-31
Sanlouise/fix scroll error

sanlouise:
## Description
We are currently using the `scrollOnShow` prop on the AlertBox in a single place, `src/platform/user/profile/vet360/components/base/Vet360EditModalErrorMessage.jsx` in `vets-website`. We now need to use the scroll functionality in other places as well, and the Alert should not scroll to the end of the viewport, instead to the start or center.

I defaulted the position to `start` instead of `end`, and will make sure to pass the `scrollPosition: end` to the one place where we are currently using the scroll logic. We'd expect most scrolling behavior to want to default to `start`, top of the screen - but this is now overridable through the `scrollPosition` prop.

[PR for frontend implementation](https://github.com/department-of-veterans-affairs/vets-website/pull/13672)

## Testing done
Works locally.

## Acceptance criteria
- [x] Add prop to AlertBox to allow for customizable scrolling position.

## Definition of done

### DIFF
--- a/packages/formation-react/src/components/AlertBox/AlertBox.jsx
+++ b/packages/formation-react/src/components/AlertBox/AlertBox.jsx
@@ -23,9 +23,9 @@ class AlertBox extends React.Component {
       return;
     }
 
-    const isInView = this._ref && window.scrollY <= this._ref.offsetTop;
+    const isInView = window.scrollY <= this._ref.offsetTop;
 
-    if (!isInView) {
+    if (this._ref && !isInView) {
       this._ref.scrollIntoView({
         block: this.props.scrollPosition,
         behavior: 'smooth',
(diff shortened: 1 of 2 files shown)


packages/formation-react/src/components/AlertBox/AlertBox.jsx
```diff
@@ -146,6 +151,7 @@ AlertBox.propTypes = {
 /* eslint-enable consistent-return */
 
 AlertBox.defaultProps = {
+  scrollPosition: 'start',
```
kelsonic, line 154: We may want to keep this as `'end'` for all the other AlertBoxes across the site that are currently expecting it to have that behavior, then in your code you can provide a custom `scrollPosition` prop with the value `'start'`

kelsonic: Oops just read the description, `'start'` works as a default prop then assuming it doesn't break any of the tests/desired behavior! 🙂

sanlouise: There is only one place where we are currently using the scroll behavior, and I made sure to pass in the correct prop in the FE PR to make sure to keep existing functionality.

Issue #362, opened 2020-08-03
[ALERT BOX]: Need to revert the hard-coded role value for screen reader usability


1Copenut:
**Describe the bug**

PR #309 added a `role="alert"` value to the Alertbox component. The VSP frontend tools team is reverting this PR because it is causing all screen readers to interrupt normal document readout to announce Alertbox values. There should be a way to add a role when instances of the Alertbox component are used, but I'd like to not use a hard-coded value.

/cc @zurbergram for visibility

**Expected behavior**

- [ ] Alert boxes should not be announced immediately unless specifically flagged for `role="alert"`

PR #363, opened 2020-08-03
Revert "add role="alert" (#309)"

cvalarida:
This reverts commit 82166c14b685817a38a9fba0663b84afb0e28abc.

## Description
Closes https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/issues/362

## Testing done
None.

## Screenshots
N/A

### DIFF
--- a/packages/formation-react/src/components/AlertBox/AlertBox.jsx
+++ b/packages/formation-react/src/components/AlertBox/AlertBox.jsx
@@ -63,7 +63,6 @@ class AlertBox extends React.Component {
         ref={ref => {
           this._ref = ref;
         }}
-        role="alert"
       >
         <div className="usa-alert-body">
           {alertHeading && <H className="usa-alert-heading">{alertHeading}</H>}
--- a/packages/formation-react/package.json
+++ b/packages/formation-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "5.4.8",
+  "version": "5.4.9",
   "description": "React implementation of patterns in Formation",
   "keywords": [
     "react",


erikphansen:
Just a heads up that pulling this into `vets-website` is going to result in test failures. Might want to loop in @kelsonic since he ran into them when the inverse of this PR was pulled into `vets-website`

cvalarida:
Ah, good to know. Thanks, @erikphansen! :bow:

kelsonic:
I think there were only 3 test files I had to update to support `role="alert"` for the `<AlertBox />` in `vets-website`:

https://github.com/department-of-veterans-affairs/vets-website/pull/13607/files#diff-c1b8424ccacefd9a6fe92872f162f55a
https://github.com/department-of-veterans-affairs/vets-website/pull/13607/files#diff-1d2a1deb0c80c4be977bde3281674562
https://github.com/department-of-veterans-affairs/vets-website/pull/13607/files#diff-f810bada69dcd84ed5fe1d35b31a2353

Hope that helps! 🙂

erikphansen:
Ohhhh... looks like those changes have nothing to do with the change to `AlertBox`. So the only tests that will break when `AlertBox` loses the `role="alert"` will be mine. So those tests of mine that will fail in the future can be commented out as a short term fix.

packages/formation-react/src/components/AlertBox/AlertBox.jsx
```diff
@@ -63,7 +63,6 @@ class AlertBox extends React.Component {
         ref={ref => {
           this._ref = ref;
         }}
-        role="alert"
```
zurbergram, line 66: Could setting `role` be done by passing in a property for the cases where an AlertBox needs a role?

erikphansen, line 66: We need to bump the version in `package.json` as well.

erikphansen:
Would be super duper if this change got merged and published (along with the corresponding change to `vets-website`) soonish. Anyone using a screen reader has been having a less-than-great experience using VA.gov :(

PR #373, opened 2020-08-21
Add notClickable prop to Telephone component

sanlouise:
## Description

In Profile 2.0, we need to render a user's phone number with the same formatting as our current `Telephone` component. However, the phone number is the user's OWN phone number, therefore it needs to not be clickable.

Instead of using css to disable the onClick of the `a` tag (which possibly could lead to a11y issues), this PR adds an optional `notClickable` prop that instead of an `a` tag renders a `span`.

I believe it's a good idea to use the `Telephone` component for this use case, albeit modified, to ensure consistency across formatting and `aria-labels`.

## Testing done
Tested locally, works as expected. Updated unit tests.

## Acceptance criteria
- [x] Pass a `notClickable` prop to the `Telephone` component to allow for purely presentational phone numbers

## Definition of done

### DIFF
--- a/packages/formation-react/src/components/Telephone/Telephone.unit.spec.jsx
+++ b/packages/formation-react/src/components/Telephone/Telephone.unit.spec.jsx
@@ -176,11 +176,8 @@ describe('Widget <Telephone />', () => {
     const wrapper = shallow(
       <Telephone contact={CONTACTS.GI_BILL} className="foo" notClickable />,
     );
-    const props = wrapper.props();
     expect(wrapper.exists('span')).to.equal(true);
     expect(wrapper.exists('a')).to.equal(false);
-    expect(props.href).to.not.exist;
-    expect(props.className).to.equal('no-wrap foo');
     expect(wrapper.text()).to.equal('888-442-4551');
     wrapper.unmount();
   });
--- a/packages/formation-react/src/components/Telephone/Telephone.unit.spec.jsx
+++ b/packages/formation-react/src/components/Telephone/Telephone.unit.spec.jsx
@@ -178,7 +178,8 @@ describe('Widget <Telephone />', () => {
     );
     expect(wrapper.exists('span')).to.equal(true);
     expect(wrapper.exists('a')).to.equal(false);
-    expect(wrapper.text()).to.equal('888-442-4551');
+    expect(wrapper.text()).to.include('888-442-4551');
+    expect(wrapper.text()).to.include('8 8 8. 4 4 2. 4 5 5 1');
     wrapper.unmount();
   });
 
--- a/packages/formation-react/src/components/Telephone/Telephone.unit.spec.jsx
+++ b/packages/formation-react/src/components/Telephone/Telephone.unit.spec.jsx
@@ -177,6 +177,8 @@ describe('Widget <Telephone />', () => {
       <Telephone contact={CONTACTS.GI_BILL} className="foo" noHref />,
     );
     const props = wrapper.props();
+    expect(wrapper.exists('span')).to.equal(true);
+    expect(wrapper.exists('a')).to.equal(false);
     expect(props.href).to.not.exist;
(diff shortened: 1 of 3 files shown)


jenstrickland:
> LGTM! I appreciate @sanlouise refactoring this to use the aria-hidden span and VA design system screen reader only span when making a phone number a non-link. We consulted over Slack on this PR, and determined it would be a better approach to remove the `href` and use the two-span approach for non-linked phone numbers. This way we can retain the easier to understand read out, and not have extra, non-functioning links in the screen reader Links menu.

Once this is in staging, can we test this in the screen readers, to make sure it works as expected?

billfienberg:
> LGTM! I appreciate @sanlouise refactoring this to use the aria-hidden span and VA design system screen reader only span when making a phone number a non-link. We consulted over Slack on this PR, and determined it would be a better approach to remove the `href` and use the two-span approach for non-linked phone numbers. This way we can retain the easier to understand read out, and not have extra, non-functioning links in the screen reader Links menu.

@1Copenut, you're saying that screen readers will still include an `<a>` element in the Links menu even when the `<a>` has no `href` attribute, correct?

1Copenut:
@billfienberg If the `notClickable` boolean prop is passed, the two spans should be rendered in stead of the `<a>` tag. I specifically wanted the link gone, because it will announce itself to screen readers when users are navigating using arrow keys, and will appear in the Links menu, but is not actionable without the `href`. It was better to have a readable text block than a disabled link, which led to Sandra's coded solution.

billfienberg:
> @billfienberg If the `notClickable` boolean prop is passed, the two spans should be rendered in stead of the `<a>` tag. I specifically wanted the link gone, because it will announce itself to screen readers when users are navigating using arrow keys, and will appear in the Links menu, but is not actionable without the `href`. It was better to have a readable text block than a disabled link, which led to Sandra's coded solution.

@1Copenut, makes sense. Thanks for the additional context.

packages/formation-react/src/components/Telephone/Telephone.unit.spec.jsx
```diff
@@ -171,6 +171,20 @@ describe('Widget <Telephone />', () => {
...
+  // notClickable
+  it('should render a span instead of an a tag', () => {
+    const wrapper = shallow(
+      <Telephone contact={CONTACTS.GI_BILL} className="foo" notClickable />,
+    );
+    const props = wrapper.props();
+    expect(wrapper.exists('span')).to.equal(true);
+    expect(wrapper.exists('a')).to.equal(false);
+    expect(props.href).to.not.exist;
+    expect(props.className).to.equal('no-wrap foo');
```
billfienberg, line 183: This assertion is failing in [Jenkins](http://jenkins.vfs.va.gov/blue/organizations/jenkins/testing%2Fveteran-facing-services-tools/detail/sanlouise-add-nonclickable-telephone/8/pipeline). It makes sense that it is failing because `foo` is being passed as the `className` prop, but the assertion is expecting to see `no-wrap foo`. 

This assertion may not even be necessary because it seems to be testing that props are passed correctly, which is something that we can reasonably trust React to do consistently.

billfienberg:
@sanlouise, the code makes sense. I'll approve the PR once the CI tests pass.

meganhkelley:
hey folks! in the future, there is a process that VSP asks folks to follow when requesting a design system change. the happy path is described [here](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/design/design-system/governance-process/edge-cases.md) (doc is WIP), and the full "design system governance process" is linked to from [here](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/projects). 

tl/dr — changes to the design system (such as this one) should be proposed to design system council for review. 

I checked with Kevin Hoffman, and this change does not need retroactive review, but I did create a ticket to reflect that our documentation will need to be updated accordingly. https://github.com/department-of-veterans-affairs/vets-design-system-documentation/projects/3#card-44220909

let me know if anyone has questions!

PR #447, opened 2020-09-29
Add ALERT_TYPE enum for use with AlertBox `status` prop

erikphansen:
I like constants 🤷 

It's always felt wrong to me to do something like:
```
<AlertBox status="info">
```

With this change we can do:
```
<AlertBox status={ALERT_TYPE.INFO}>
```

Why? Better code completion from editors and/or earlier error/typo detection. It's not TypeScript, but it's better than throwing strings around.


sanlouise (approved):
I love it! 👍

bkjohnson (approved):

PR #452, opened 2020-10-01
Fix alert box scroll when scrollOnShow is true

sanlouise:
## Description
We also need to scroll to an AlertBox when `scrollOnShow` is passed and the AlertBox is already (partially) in the viewport). Initially, the scroll logic would not run when the AlertBox was only a few pixels in the viewport which was confusing for the user.

I needed to use a `setTimeout` to ensure the `ref` in the scroll logic was defined, since putting any logic in `componentDidMount` does not guarantee that it's actually already painted to the DOM.

## Testing done
Works locally.

## Acceptance criteria
- [x] Also scroll when the AlertBox is already in the viewport

## Definition of done

### DIFF
--- a/packages/formation-react/src/components/AlertBox/AlertBox.jsx
+++ b/packages/formation-react/src/components/AlertBox/AlertBox.jsx
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { Component } from 'react';
 import classNames from 'classnames';
 
 // Enum used to set the AlertBox's `status` prop
@@ -9,11 +9,15 @@ export const ALERT_TYPE = Object.freeze({
   SUCCESS: 'success', // Green border, green checkmark
   WARNING: 'warning', // Yellow border, black triangle exclamation
   CONTINUE: 'continue', // Green border, green lock
-})
+});
 
-class AlertBox extends React.Component {
+class AlertBox extends Component {
   componentDidMount() {
     this.scrollToAlert();
+  }
+
+  componentWillUnmount() {
+    clearTimeout(this.scrollToAlertTimeout);
   }
 
   shouldComponentUpdate(nextProps) {
@@ -32,13 +36,14 @@ class AlertBox extends React.Component {
       return;
     }
 
-    const isInView = window.scrollY <= this._ref.offsetTop;
-
-    if (this._ref && !isInView) {
-      this._ref.scrollIntoView({
-        block: this.props.scrollPosition,
-        behavior: 'smooth',
-      });
+    if (this._ref) {
+      clearTimeout(this.scrollToAlertTimeout);
+      this.scrollToAlertTimeout = setTimeout(() => {
+        this._ref.scrollIntoView({
+          block: this.props.scrollPosition,
+          behavior: 'smooth',
+        });
+      }, 0);
     }
   };
 
--- a/packages/formation-react/src/components/AlertBox/AlertBox.jsx
+++ b/packages/formation-react/src/components/AlertBox/AlertBox.jsx
@@ -32,14 +32,12 @@ class AlertBox extends Component {
   }
 
   scrollToAlert = () => {
-    if (!this.props.isVisible || !this.props.scrollOnShow) {
-      return;
-    }
-
-    if (this._ref) {
+    // Without using the setTimeout, React has not added the element
+    // to the DOM when it calls scrollIntoView()
+    if (this.props.isVisible && this.props.scrollOnShow) {
       clearTimeout(this.scrollToAlertTimeout);
       this.scrollToAlertTimeout = setTimeout(() => {
-        this._ref.scrollIntoView({
+        this._ref?.scrollIntoView({
           block: this.props.scrollPosition,
           behavior: 'smooth',
         });
--- a/packages/formation-react/package.json
+++ b/packages/formation-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "5.6.1",
+  "version": "5.7.0",
   "description": "React implementation of patterns in Formation",
   "keywords": [
     "react",


erikphansen:
So in the end you didn't need to add any "is partially in the viewport" logic/the problem had nothing to do with the AlertBox already being partially in view?

kelsonic (approved):
LGTM! 🎉 Thanks for clearing the timeout on componentWillUnmount too!

packages/formation-react/src/components/AlertBox/AlertBox.jsx
```diff
@@ -32,13 +36,14 @@ class AlertBox extends React.Component {
...
-        behavior: 'smooth',
-      });
+    if (this._ref) {
+      clearTimeout(this.scrollToAlertTimeout);
+      this.scrollToAlertTimeout = setTimeout(() => {
+        this._ref.scrollIntoView({
+          block: this.props.scrollPosition,
+          behavior: 'smooth',
+        });
+      }, 0);
```
bkjohnson, line 46: > I needed to use a `setTimeout` to ensure the ref in the scroll logic was defined

Can you elaborate on this? I apologize if I'm missing something obvious - it just isn't clear why we have `setTimeout` with a delay of 0 along with a conditional check for the ref.

Would any of this be simplified if this component could use a [hook-based ref](https://reactjs.org/docs/hooks-reference.html#useref)?

sanlouise: @bkjohnson I needed the `ref` to get in the event loop because it wouldn't point to the DOM node otherwise - since it's not painted yet. I believe using a hook-based ref would result in more or less the same logic 🙂

bkjohnson: React's [docs on `refs`](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs) say that:

> Refs are guaranteed to be up-to-date before `componentDidMount` or `componentDidUpdate` fires.

Is there still something missing?

sanlouise: Correct, I read that, too. However, if you try it without wrapping it in a setTimeout, [this line](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/452/files#diff-4c7206ecbf59396b33d877282274b633R39) equates to false because `this._ref` is undefined. I have also experimented with using `React.createRef` and the same issue occurs where the ref isn't set yet when `scrollToAlert` is called within componentDidMount. Feel free to pull down the branch and you will see what I mean

bkjohnson: I used `yarn link` to connect this branch to `vets-website`, and it doesn't look like `this._ref` is undefined in any of the unit tests:

image


image

Can you show me where it is appearing as undefined?

bkjohnson (commented):
Ignoring the timeout and ref for a moment, do you think it would be possible to refactor things to get this? I feel like this makes the intention of the props more explicit.

```diff
diff --git a/packages/formation-react/src/components/AlertBox/AlertBox.jsx b/packages/formation-react/src/components/AlertBox/AlertBox.jsx
index 59e54223f..29623cf43 100644
--- a/packages/formation-react/src/components/AlertBox/AlertBox.jsx
+++ b/packages/formation-react/src/components/AlertBox/AlertBox.jsx
@@ -32,14 +32,11 @@ class AlertBox extends Component {
   }
 
   scrollToAlert = () => {
-    if (!this.props.isVisible || !this.props.scrollOnShow) {
-      return;
-    }
 
-    if (this._ref) {
+    if (this.props.isVisible && this.props.scrollOnShow) {
       clearTimeout(this.scrollToAlertTimeout);
       this.scrollToAlertTimeout = setTimeout(() => {
-        this._ref.scrollIntoView({
+        this._ref?.scrollIntoView({
           block: this.props.scrollPosition,
           behavior: 'smooth',
         });
```

packages/formation-react/src/components/AlertBox/AlertBox.jsx
erikphansen: I took a look at this locally as well. And what I found was that checking for `this._ref` in `scrollToAlert` was not required, but putting the `scrollIntoView()` call in a timeout was. This GIF shows the Alert not scrolling into view when the `scrollIntoView()` call is outside of a timeout and then it successfully scrolling into view when the call is made in a timeout. I'm very perplexed why this is needed.

without and with timeout

Screen Shot 2020-10-06 at 3 40 54 PM

In summary: the `this_ref` check does not seem needed. But delaying the `scrollIntoView()` call until the next run of the event loop seems to be needed 🤔

sanlouise:
Thanks for the suggestion @bkjohnson! I would like to implement it but unfortunately it looks like this repo doesn't support `?.`, which means we would need to use an if statement there or within the setTimeout to check for `this._ref` to prevent a runtime error 🙂

packages/formation-react/src/components/AlertBox/AlertBox.jsx
sanlouise: So @erikphansen and I were testing this and it looks like I was mistaken. `this._ref` is not `undefined`, but it does not seem to scroll without putting it in the event loop (likely due to the dom node not being fully painted at the point of execution, just a hypothesis though). Here is a screenshot from Erik:

image (9)

bkjohnson:
@sanlouise, if you merge in master you should be able to use optional chaining now!

packages/formation-react/src/components/AlertBox/AlertBox.jsx
bkjohnson: This is strange, but I think it's ok for now since an alternative would likely involve some out of scope refactoring. Can you add some comments around the `clearTimeout` and `setTimeout` calls to explain why they are there?

erikphansen: Amen on this being strange. But I'm with Sandra. I don't think React has actually added the element to the DOM when it calls `scrollIntoView()`.

erikphansen: I also see no reason to store the timeout and clean it up since it essentially executes immediately.

sanlouise: I try to follow the best practices when it comes to using `setTimeout` and `setInterval` to always clear them in order to prevent memory leaks. In this scenario, the probability is very low (if not 0) due to the setTimeout time being set to 0, but I like to play it safe anyways in case someone were to change that number in the future

sanlouise: @bkjohnson I added a comment to the setTimeout, good idea!

bkjohnson (approved):

PR #486, opened 2020-10-29
15408: Storybook Component - ErrorableCheckbox

craigwheeler:
## Description
Created story for ErrorableCheckbox component | closes [#15408](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/15408)

## Screenshots
Library  ErrorableCheckbox - Default ⋅ Storybook 2020-10-30 09-03-44
Library  ErrorableCheckbox - Default ⋅ Storybook 2020-10-30 09-03-23

## Acceptance criteria
- [x] All the "stories" found in `ErrorableCheckbox.mdx` are written as Storybook stories
- [x] These stories are found in `packages/formation-react/src/components/ErrorableCheckbox/ErrorableCheckbox.stories.jsx`

## Definition of done


packages/formation-react/src/components/ErrorableCheckbox/ErrorableCheckbox.stories.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import ErrorableCheckbox from './ErrorableCheckbox';
+
+export default {
+  title: 'Library/ErrorableCheckbox',
+  component: ErrorableCheckbox,
+  argTypes: {
+    required: { control: 'boolean' },
+    errorMessage: { control: 'text' },
+    id: { control: 'text' },
```
cvalarida, line 10: Looks like this isn't a prop.

packages/formation-react/src/components/ErrorableCheckbox/ErrorableCheckbox.stories.jsx
```diff
@@ -0,0 +1,33 @@
...
+    title: { control: 'text' },
+    label: { control: 'text' },
+    labelAboveCheckbox: { control: 'text' },
+  },
```
cvalarida, line 14: I think the only ones you'll need here are `label` and `ariaLabelledBy` since those are functions in the props. The rest should be gotten automagically from the `propTypes` on the component.

craigwheeler: Actually I was able to remove the `argTypes` from this one as well since `defaultArgs` is passing the default content vals and the rest of the props can be inferred

packages/formation-react/src/components/ErrorableCheckbox/ErrorableCheckbox.stories.jsx
```diff
@@ -0,0 +1,23 @@
...
+export default {
+  title: 'Library/ErrorableCheckbox',
+  component: ErrorableCheckbox,
+};
+
+const Template = (args) => <ErrorableCheckbox {...args} />;
+
+const defaultArgs = {
+  required: true,
+  errorMessage: 'Error message',
```
cvalarida, line 13: What do you think about making the default without an error message and then making another story with the error state?

cvalarida (approved):
It doesn't render quite right right now. I had to add some padding to the canvas' root so I could see the red line on the left, but after I did that, it worked fine. I'll make a separate PR to add the padding.
image

cvalarida:
:thinking: Actually, after looking into it a little, I think it might be better to wrap the contents of `Template` in a `<div style={{ paddingLeft: '1em' }}>` or something.

What do you think?

packages/formation-react/src/components/ErrorableCheckbox/ErrorableCheckbox.stories.jsx
craigwheeler: ahh yea good idea the default should be the primary state ill update it tks

craigwheeler:
> 🤔 Actually, after looking into it a little, I think it might be better to wrap the contents of `Template` in a `<div style={{ paddingLeft: '1em' }}>`.
> image
> 
> What do you think?

---

definitely ... I've actually noticed the same thing with other components and wondered if maybe these should be rendered in some kind of container so my thoughts were `padding` or `max-width` container which would give us a consistent 5% margin on either side of the component? 
`  <div style={{ maxWidth: '90%', margin: '0 auto' }}>`

cvalarida:
> my thoughts were `padding` or `max-width` container which would give us a consistent 5% margin on either side of the component

I was thinking that'd be a good idea too, but then realized it'd give us an unfaithful representation while in mobile view (see screenshot). I'm not sure how much of a problem that'd be, but for now, I think the answer in this case is to add the padding to this one component instead of all components.

image

PR #494, opened 2020-11-03
Fix for missing percentage sign in the flex-basis property with a value of zero

ncksllvn:
## Description
The `flex-basis` property should always be expressed as a percentage. However, in the `flex` property shorthand, a value of zero is outputted without a percentage sign. Ultimately, this PR changes `flex: 1 0 0` to output as `flex: 1 0 0%`.

https://github.com/department-of-veterans-affairs/va.gov-team/issues/15610

## Testing done
I looked at the output in formation.min.css

image

### issue in IE
notice the red squiggly

and that the property doesn't work. The input should be inline with the button.

## Acceptance criteria
- [ ] `flex-basis` is outputted properly

## Definition of done


packages/formation/sass/utilities/_flexbox.scss
```diff
@@ -2,11 +2,11 @@
...
-  2 0 0,
-  3 0 0,
-  4 0 0,
-  1 0 0,
+  1 0 0%,
+  2 0 0%,
+  3 0 0%,
+  4 0 0%,
+  1 0 0%,
   0 0 auto
```
bkjohnson, line 10: I had to do a bit of learnin' for this. According to [the docs](https://developer.mozilla.org/en-US/docs/Web/CSS/flex), in the three-value syntax, the third value is the width for the `flex-basis`.

Additionally:

image


So, the unit (`px`, `%`, etc.) is required.

bkjohnson (approved):
Looks good!

packages/formation/sass/utilities/_flexbox.scss
ncksllvn: Thanks for looking that up! It might be worth sweeping through for other instances where this could happen. You would think that a `0` is `0` no matter what the unit is, but 🤷

PR #498, opened 2020-11-04
Move react-scroll to formation-react

bkjohnson:
## Description

The `react-scroll` dependency is declared in the root `package.json`, but only ever used in `formation-react`. This lack of a dependency causes problems for us when we try to deploy storybook from https://github.com/department-of-veterans-affairs/component-library

This change will declare the dependency in the child package (formation-react), so that when we update the mirrored repo, the storybook deploy shouldn't run into any problems.

## Screenshots

image

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,6 @@
     "react-docgen": "^2.19.0",
     "react-dom": "^16.7.0",
     "react-router": "3",
-    "react-scroll": "^1.7.7",
     "react-test-renderer": "^16.7.0",
     "recast": "^0.14.4",
     "resolve-url-loader": "^2.2.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -91,7 +91,6 @@
     "react-docgen": "^2.19.0",
     "react-dom": "^16.7.0",
     "react-router": "3",
-    "react-scroll": "^1.7.7",
```
cvalarida, line 94: Did you confirm no other packages used this?

bkjohnson: In the screenshot you can see that it is only found in the `formation-react` package.

cvalarida: Ahh, I totally overlooked the screenshot. :see_no_evil: Good call. :+1:

cvalarida (approved):
:shipit:

U-DON (approved):

PR #500, opened 2020-11-05
Miscellaneous updates to formation-react

cvalarida:
## Description
Just a few minor quality-of-life improvements.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,6 @@ lerna-debug.log
 node_modules
 npm-debug.log
 package-lock.json
-packages/formation-react/*.js
 public/dist
 yarn-error.log
 .env
--- a/packages/formation-react/.gitignore
+++ b/packages/formation-react/.gitignore
@@ -1,1 +1,2 @@
 yarn.lock
+*.js
--- a/packages/formation-react/.gitignore
+++ b/packages/formation-react/.gitignore
@@ -1,2 +1,2 @@
 yarn.lock
-*.js
+/*.js
--- a/packages/formation-react/README.md
+++ b/packages/formation-react/README.md
@@ -13,9 +13,8 @@ See [design system](https://department-of-veterans-affairs.github.io/design-syst
 - AdditionalInfo.js
 - AlertBox.js
 - Breadcrumbs.js
-- ExpandingGroup.js
-- ErrorableDate.js
 - ErrorableCheckbox.js
+- ErrorableDate.js
 - ErrorableFileInput.js
 - ErrorableMonthYear.js
 - ErrorableNumberInput.js
@@ -23,11 +22,13 @@ See [design system](https://department-of-veterans-affairs.github.io/design-syst
 - ErrorableSelect.js
 - ErrorableTextArea.js
 - ErrorableTextInput.js
+- ExpandingGroup.js
 - IconBase.js
 - IconHelp.js
 - IconSearch.js
 - IconUser.js
 - LoadingIndicator.js
+- MegaMenu.js
 - Modal.js
 - OMBInfo.js
 - Pagination.js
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -12,7 +12,6 @@ lerna-debug.log
 node_modules
 npm-debug.log
 package-lock.json
-packages/formation-react/*.js
```
cvalarida, line 15: Added this to `formation-react`'s own `.gitignore`.

packages/formation-react/README.md
```diff
@@ -13,21 +13,22 @@ See [design system](https://department-of-veterans-affairs.github.io/design-syst
 - AdditionalInfo.js
 - AlertBox.js
 - Breadcrumbs.js
-- ExpandingGroup.js
```
cvalarida, line 16: Sorted this list.

packages/formation-react/scripts/build.js
```diff
@@ -49,13 +49,13 @@ const fileNames = [].concat.apply(
   [],
   [
     glob.sync('./src/components/**/*.jsx', {
-      ignore: ['./**/*.unit.spec.jsx', './**/*.stories.(mdx|jsx)'],
+      ignore: ['./**/*.unit.spec.jsx', './**/*.stories.jsx'],
```
cvalarida, line 52: Stopped `yarn build` from building stories. (Which I thought I had already done, but apparently failed.)

packages/formation-react/src/components/ErrorableCheckboxGroup/ErrorableCheckboxGroup.stories.jsx
```diff
@@ -11,7 +11,11 @@ export default {
   },
 };
 
-const Template = (args) => <ErrorableCheckboxGroup {...args} />;
+const Template = (args) => (
+  <div style={{ paddingLeft: '1em' }}>
+    <ErrorableCheckboxGroup {...args} />
+  </div>
+);
```
cvalarida, line 18: Added some padding to the left so we can see the red bar for the error.

packages/formation-react/src/components/ErrorableFileInput/ErrorableFileInput.stories.jsx
```diff
@@ -6,7 +6,11 @@ export default {
   component: ErrorableFileInput,
 };
 
-const Template = (args) => <ErrorableFileInput {...args} />;
+const Template = (args) => (
+  <div style={{ paddingLeft: '1em' }}>
+    <ErrorableFileInput {...args} />
+  </div>
+);
```
cvalarida, line 13: Same here.

bkjohnson (approved):
Looks good! I wonder if at some point we could/should do something with `preview-body` to give everything more of an inset so that we don't have to have those custom styled wrappers, but that's for another day :upside_down_face:

U-DON (approved):

packages/formation-react/src/components/ErrorableCheckboxGroup/ErrorableCheckboxGroup.stories.jsx
U-DON: I usually opt for `rem` units, but I don't know if we have a standardized unit (and maybe we should).

Clark-W-Griswold (approved):
👍

packages/formation-react/.gitignore
```diff
@@ -1 +1,2 @@
 yarn.lock
+*.js
```
U-DON, line 2: Does this capture the `src/helpers` as well? Those are `.js` files.

U-DON: Maybe it should be something like `/*.js`?

cvalarida: Ooh, good thinking. I just verified this only captures root-level `.js` files by making a change to `helpers/fields.js` and it looks like it worked as expected; the change was tracked.
image

However, _new_ `.js` files didn't get tracked, so changing it to `/*.js` is the way to go. Thanks! :raised_hands:

PR #503, opened 2020-11-05
Added logic to allow for right aligning

cohnjesse:
## Description
[Original Ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/15484)

Often times numerical data is supposed to be right aligned in tables and in fact, our design system calls for right aligning numerical data in tables as well as material design and AP style guides call for right aligning numerical data in tables as well. This PR adds a prop to the `<Table />` component to allow a key of `rightAlign` to be passed in with he value true on a field and have that field get the standard design system right align class added to right-align it.

The class that is added is from medium screens and up since on mobile the table stacks and we don't want the data right aligned then.

## Testing done
All existing unit tests pass, further unit tests are coming in future iterations

## Screenshots

Screen Shot 2020-11-05 at 2 12 19 PM

## Acceptance criteria
- [x] We are able to right-align data in columns in our `<Table />` component
- [x] A PR has been submitted for review

### DIFF
--- a/packages/formation-react/src/components/Table/Table.jsx
+++ b/packages/formation-react/src/components/Table/Table.jsx
@@ -8,12 +8,13 @@ const rowPaddingClass = 'vads-u-padding-y--2';
 
 function Table(props) {
   const { currentSort, fields, data, ariaLabelledBy } = props;
+  let alignClass;
 
   return (
     <table aria-labelledby={ariaLabelledBy} className="responsive" role="table">
       <thead>
         <tr role="row">
-          {fields.map((field) =>
+          {fields.map(field =>
             field.nonSortable ? (
               <th key={field.value}>{field.label}</th>
             ) : (
@@ -48,19 +49,26 @@ function Table(props) {
             className={`${borderClasses} ${rowPaddingClass}`}
             role="row"
           >
-            {fields.map((field, index) => (
-              <td
-                data-index={index}
-                className={classNames(borderClasses, {
-                  'vads-u-text-align--left': field.alignLeft,
-                })}
-                data-label={`${field.label}:`}
-                key={`${rowIndex}-${field.label}`}
-                role="cell"
-              >
-                {item[field.value] === null ? '---' : item[field.value]}
-              </td>
-            ))}
+            {fields.map((field, index) => {
+              if (field.alignRight) {
+                alignClass = 'vads-u-text-align--right';
+              } else if (field.alignLight) {
+                alignClass = 'vads-u-text-align--left';
+              } else {
+                alignClass = 'vads-u-text-align--left';
+              }
+              return (
+                <td
+                  data-index={index}
+                  className={classNames(borderClasses, alignClass)}
+                  data-label={`${field.label}:`}
+                  key={`${rowIndex}-${field.label}`}
+                  role="cell"
+                >
+                  {item[field.value] === null ? '---' : item[field.value]}
+                </td>
+              );
+            })}
           </tr>
         ))}
       </tbody>
@@ -88,6 +96,7 @@ Table.propTypes = {
       value: PropTypes.string,
       nonSortable: PropTypes.boolean,
       alignLeft: PropTypes.boolean,
+      alignRight: PropTypes.boolean,
     }),
   ),
   /**
--- a/packages/formation-react/src/components/Table/Table.jsx
+++ b/packages/formation-react/src/components/Table/Table.jsx
@@ -51,15 +51,10 @@ function Table(props) {
             {fields.map((field, index) => (
               <td
                 data-index={index}
-                className={classNames(
-                  borderClasses,
-                  {
-                    'vads-u-text-align--left': field.alignLeft,
-                  },
-                  {
-                    'medium-screen:vads-u-text-align--right': field.alignRight,
-                  },
-                )}
+                className={classNames(borderClasses, {
+                  'vads-u-text-align--left': field.alignLeft,
+                  'medium-screen:vads-u-text-align--right': field.alignRight,
+                })}
                 data-label={`${field.label}:`}
                 key={`${rowIndex}-${field.label}`}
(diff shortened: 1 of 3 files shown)


packages/formation-react/src/components/Table/Table.jsx
```diff
@@ -48,19 +49,26 @@ function Table(props) {
             className={`${borderClasses} ${rowPaddingClass}`}
             role="row"
           >
-            {fields.map((field, index) => (
-              <td
-                data-index={index}
-                className={classNames(borderClasses, {
-                  'vads-u-text-align--left': field.alignLeft,
```
bkjohnson, line 55: Could this be accomplished by adding another entry here?

```
'medium-screen:vads-u-text-align--right': field.alignRight
```

packages/formation-react/src/components/Table/Table.jsx
```diff
@@ -48,19 +49,26 @@ function Table(props) {
...
-                {item[field.value] === null ? '---' : item[field.value]}
-              </td>
-            ))}
+            {fields.map((field, index) => {
+              if (field.alignRight) {
+                alignClass = 'medium-screen:vads-u-text-align--right';
+              } else if (field.alignLight) {
+                alignClass = 'vads-u-text-align--left';
+              } else {
+                alignClass = 'vads-u-text-align--left';
```
bkjohnson, line 58: This `else` does the same thing as the `else if` above it. If my previous comment isn't feasible to implement, then we should change how this if statement is set up.

bkjohnson (changes requested):
In addition to the functional change, can you try adding a Storybook story to demonstrate this? You can see the current ones [here](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/master/packages/formation-react/src/components/Table/Table.stories.jsx). To run storybook locally, just run `yarn storybook`.

packages/formation-react/src/components/Table/Table.jsx
cohnjesse: @bkjohnson Yep it can, I figured putting it in an if statement would be easier to read but thinking about it I suppose it is really not. I updated this.

packages/formation-react/src/components/Table/Table.jsx
cohnjesse: @bkjohnson Yep, I updated this to be inline using `classNames`

packages/formation-react/src/components/Table/Table.jsx
```diff
@@ -51,9 +51,15 @@ function Table(props) {
...
-                })}
+                className={classNames(
+                  borderClasses,
+                  {
+                    'vads-u-text-align--left': field.alignLeft,
+                  },
+                  {
+                    'medium-screen:vads-u-text-align--right': field.alignRight,
+                  },
+                )}
```
bkjohnson, line 62: These can be combined into the same object which reads easier in my opinion.

An [example based on the docs](https://github.com/JedWatson/classnames#usage):

```js
const borderClasses = 'test';

classNames(borderClasses,
    { 
        foo: true
    }, 
    { 
        bar: true
    }
); // => 'test foo bar'

classNames(borderClasses,
    { 
        foo: true, 
        bar: true 
    }
); // => 'test foo bar'
```

bkjohnson (commented):
Let us know if you have any questions about storybook or how to add your story.

cohnjesse:
@bkjohnson I tried running storybook locally inside `veteran-facing-services-tools` and yarn doesn't have it as a command

packages/formation-react/src/components/Table/Table.jsx
cohnjesse: Yep, updated this

packages/formation-react/src/components/Table/Table.stories.jsx
```diff
@@ -109,6 +109,19 @@ AlignLeft.args = {
...
+alignRight.args = {
+  ...defaultArgs,
+  data: [
+    {
+      title: 'alignRight',
+      description:
+        'If you need to right align data in a column, in the fields object you can set alignRight to true for that field. This will right align the entire column which is generally used for numerical data in a table.',
+    },
+  ],
+};
```
bkjohnson, line 123: This doesn't actually demonstrate anything being aligned to the right:

image


I think this would do a better job:

```diff
diff --git a/packages/formation-react/src/components/Table/Table.stories.jsx b/packages/formation-react/src/components/Table/Table.stories.jsx
index 88befaac7..67574c769 100644
--- a/packages/formation-react/src/components/Table/Table.stories.jsx
+++ b/packages/formation-react/src/components/Table/Table.stories.jsx
@@ -113,11 +113,13 @@ AlignLeft.args = {
 export const alignRight = Template.bind({});
 alignRight.args = {
   ...defaultArgs,
-  data: [
+  fields: [
+    defaultArgs.fields[0],
+    defaultArgs.fields[1],
     {
-      title: 'alignRight',
-      description:
-        'If you need to right align data in a column, in the fields object you can set alignRight to true for that field. This will right align the entire column which is generally used for numerical data in a table.',
+      label: 'Year',
+      value: 'year',
+      alignRight: true,
     },
   ],
 };
```

With `alignRight`:

image


Without `alignRight`:

image

bkjohnson (commented):
Once the story is fixed this should be good to go!

Also, don't forget to update the minor version number in the `formation-react` package.json so that this can get published.

packages/formation-react/src/components/Table/Table.stories.jsx
cohnjesse: Yeah, I totally misunderstood the idea of a story LOL, just fixed this

bkjohnson (approved):
Looks good! Just a reminder that you won't be able to get this in `vets-website` yet without updating the version number.

U-DON (approved):

PR #543, opened 2020-11-20
Check branch before publishing

cvalarida:
## Description
Like it says on the tin, this PR adds a `prepublishOnly` script to check if the branch is `master` to prevent publishing from non-master branches (which has happened a few times now).

## Testing done
See screenshots.

## Screenshots
Test that the prepublish script exits with a status code 1 for non-master branches, but status code 0 on `master`.
image
**Note:** I updated the script name to `prepublish.js` after taking this screenshot.

Test that the `prepublishOnly` hook disallows publishing when not on `master`.
image

## Acceptance criteria
- [x] Attempting to publish from `master` is successful
- [x] Attempting to publish from a branch other than `master` fails with a warning message

### DIFF
--- a/scripts/prepublish.js
+++ b/scripts/prepublish.js
@@ -1,0 +1,21 @@
+/**
+ * Ensure that we're publishing from the master branch.
+ */
+
+const simpleGit = require('simple-git');
+const chalk = require('chalk');
+
+const git = simpleGit();
+
+(async () => {
+  await git.init();
+  const branchName = (await git.branch()).current;
+
+  if (branchName !== 'master') {
+    // eslint-disable-next-line no-console
+    console.log(
+      chalk.yellow('Please check out the master branch before publishing.\n'),
+    );
+    process.exit(1);
+  }
+})();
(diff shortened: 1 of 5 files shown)


scripts/prepublish.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+const simpleGit = require('simple-git');
+const chalk = require('chalk');
+
+const git = simpleGit();
+
+(async () => {
+  await git.init();
+  const branchName = (await git.branch()).current;
```
U-DON, line 12: Unless `simple-git` is expected to be used in more places, I think this could be accomplished with a:

```js
execSync('git rev-parse --abbrev-ref HEAD').toString().trim()
```

Don't feel too strongly either way, though.

U-DON (approved):

scripts/prepublish.js
cvalarida: Ah, fair point. I was _this_ close to just writing another bash script, but refrained myself. This seems like a good compromise. :+1:

bkjohnson (approved):

PR #547, opened 2020-11-26
Deploy Storybook to S3 using CircleCI

bkjohnson:
## Description

Part of https://github.com/department-of-veterans-affairs/va.gov-team/issues/15169

This follows the [Design section of the appropriate RFC](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/request-for-comment/2020-11-4-storybook-hosting.md#design).

This will deploy storybook to the dev & staging buckets.

(The commit history is a disaster :see_no_evil: )

- https://dev-design.va.gov/storybook/
- https://staging-design.va.gov/storybook/

## Testing done

Lots of testing in CircleCI & the browser

- Published storybook to dev/staging by triggering the [appropriate workflow](https://app.circleci.com/pipelines/github/department-of-veterans-affairs/veteran-facing-services-tools?branch=circleci-project-setup)

## Acceptance criteria
- [ ]

## Definition of done


.circleci/config.yml
```diff
@@ -0,0 +1,120 @@
...
+      - aws-cli/setup
+      - run: yarn install
+      - run:
+          name: "Build Storybook"
+          command: cd packages/formation-react && yarn build-storybook
+      - write-build-info:
+          dest: "packages/formation-react/storybook-static"
+      - copy-to-s3:
+          source: "./packages/formation-react/storybook-static"
+          bucket: "<< parameters.bucket >>/storybook"
```
bkjohnson, line 92: This job will run in the workflow if we've determined that we need to publish storybook. It uses the [CircleCI `aws-cli` orb](https://circleci.com/developer/orbs/orb/circleci/aws-cli), and the `context` in the workflow is set up with environment variables that the orb can use for the key & secret key.

This does a bit of initial setup, then the meat of the job is:

1. Build storybook
1. Write a `BUILD.txt` file to the storybook static site directory
1. Copy all of the files in the storybook static site directory to the S3 bucket, making sure that they are publicly readable.

bkjohnson (commented):
The CircleCI config is set up so that the jobs can use some commands defined in the same file.

.circleci/config.yml
```diff
@@ -0,0 +1,120 @@
...
+        description: "The destination of the build file"
+        type: string
+    steps:
+      - run:
+          name: "Writing build info"
+          command: |
+            echo "<< parameters.data >>" > << parameters.dest >>/BUILD.txt
+
+  copy-to-s3:
+    description: "Copies adirectory to a bucket (and folder) in S3"
```
cvalarida, line 38:
```suggestion
    description: "Copies a directory to a bucket (and folder) in S3"
```

.circleci/config.yml
```diff
@@ -0,0 +1,120 @@
...
+      dest:
+        description: "The destination of the build file"
+        type: string
+    steps:
+      - run:
+          name: "Writing build info"
+          command: |
+            echo "<< parameters.data >>" > << parameters.dest >>/BUILD.txt
+
+  copy-to-s3:
```
cvalarida, line 37: There's also [this orb](https://circleci.com/developer/orbs/orb/circleci/aws-s3). I'm pretty ignorant here, though. Is it better to use an orb like this or run a simple `aws s3` command and assume Circle has all the context it needs?

@dginther?

.circleci/config.yml
```diff
@@ -0,0 +1,120 @@
...
+            echo "export LAST_PUBLISHED_COMMIT=$(curl -s -f https://<< parameters.doc_site >>/storybook/BUILD.txt | sed -r 's/^.*=(.*)$/\1/')" >> $BASH_ENV
+            source $BASH_ENV
+        # If we didn't find a last published commit, then we need to publish
+      - run: |
+          if [ "$LAST_PUBLISHED_COMMIT" = "" ]; then
+            circleci-agent step halt
+          fi
+      - check-for-changes:
+          in: "formation-react"
+          since: $LAST_PUBLISHED_COMMIT
```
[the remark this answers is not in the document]
cvalarida: This makes sense to me now, but can we explain this directly in the config file so our future selves—and others—will know what the reasoning is later?

I feel like this would have saved us a lot of headache for the Metalsmith build, so I'd like to learn from that and document directly in the code here. I find commenting more on the "why" is typically more helpful than the "what," but given that this is relatively esoteric (from an FE's perspective, that is), it's probably useful to leave comments on what this is doing in here too.

cvalarida (commented):
This looks good to me, but I'm at the end of my workday and admit to not following it completely. I'll loop back around to it later (tomorrow, I hope), but I'd also like to get a review from ops to make sure we're not doing anything silly here.

@jhouse-solvd Who would be the right person to ask for a review on this CircleCI configuration PR to double-check that we're not doing anything dumb / in a weird way?

.circleci/config.yml
bkjohnson: I did see that orb when I was looking at available options, but I haven't tried it.  I'm not _against_ trying it out though. The reason I went with the `aws-cli` orb is because it allowed for less abstraction, since I could have Circle run the exact same command that I (or someone else) would have run on the command line.

cvalarida (approved):
Looks great! Thanks for adding that comment. :trophy:

U-DON (approved):

PR #551, opened 2020-12-03
16615 add standard ga event

cohnjesse:
## Description
[Original Ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/16615)

This PR is to add a standardized Google Analytics event for when a radio button is changed in a wizard page. I added the event in a way that is works based off of a prop, if that prop is not passed in then the standardized event does not fire. This is to make sure that if teams already added events of their own the events will not double up and the standardized event will not fire until they update their wizard code.

## Testing done
tested using dataslayer locally

## Screenshots
Screen Shot 2020-12-03 at 2 06 50 PM

## Acceptance criteria
- [x] Events for radio button change is now a standardized event
- [x] A PR has been requested for merging


cvalarida (approved):
cc @jonwehausen for visibility.

cvalarida (changes requested):
Oh, hey. Didn't realize what component this was on.

packages/formation-react/src/components/ErrorableRadioButtons/ErrorableRadioButtons.jsx
```diff
@@ -48,7 +47,16 @@ class ErrorableRadioButtons extends React.Component {
...
-  handleChange(domEvent) {
+  handleChange(domEvent, optionLabel) {
+    // If a Google Analytics prop was passed in, fire the standardized event onChange
+    if (this.props.standardGAEvent && window.dataLayer) {
+      window.dataLayer.push({
+        event: `howToWizard-formChange`,
+        'form-field-type': 'form-radio-buttons',
+        'form-field-label': this.props.label,
+        'form-field-value': optionLabel,
+      });
```
cvalarida, line 58: Adding an event in here is fine, but this hard-coded `howToWizard-formChange` event is a no go. Instead, pass a function prop to `ErrorableRadioButtons` which can be called with the data required to return the whole event (`optionLabel` in this case).

It'd look something like this:
```jsx
<ErrorableRadioButton
  onChangeGAEvent={({ optionLabel }) => ({
    event: 'howToWizard',
    ...
    'form-field-value': optionLabel
  })}/>
  ```

Comment (cohnjesse) [reply packages/formation-react/src/components/ErrorableRadioButtons/ErrorableRadioButtons.jsx:58]: @cvalarida So the idea with these standard events is to make it so that developers don't need to code events at all since the GA team wants a standard event always fired by this component when used in wizards.. Adding it as a function prop is not much better in this regard than firing a custom event. What about the idea of pre-fixing the `standardGAEvent` prop with where the buttons are being used? So like if they are used in a wizard the prop would be `wizardStandardGAEvent` that would only fire the event included in this PR. Then if someone doesn't add that prop then the event never fires. Thoughts?

Comment (cohnjesse) [reply packages/formation-react/src/components/ErrorableRadioButtons/ErrorableRadioButtons.jsx:58]: @cvalarida Not sure if you saw my above comment?

Comment (cvalarida) [reply packages/formation-react/src/components/ErrorableRadioButtons/ErrorableRadioButtons.jsx]: I definitely didn't! Totally missed the notification. Sorry about that. :grimacing: 

So my main concern here is that you've hardcoded the GA event `howToWizard-formChange` into _all_ radio buttons, regardless of how they're actually being used. If we decided to use a `RadioButton` (as they're now called in the latest `master`) outside of a "how to wizard," this event name no longer makes sense.

I understand the desire to cut down on boilerplate, but we need _some_ level of customization here. We can't hard code that event name.

Honestly, I'm still on the fence about adding analytics into the components at all, but I don't have a great mental model to help me determine what kind of issues we'll encounter there. I don't see an immediate problem with it.

All that said, I'd recommend building this such that the component accepts a prop which can either be a string for the event name or a function which returns the whole event. This gives folks using the component the choice of whether to cut down on boilerplate (use the string) or be really specific about what the event contains (use the function). When this new prop is absent, no GA event will be captured.

---

As a final postscript here, I wanted to mention a couple housekeeping items:
1. We've renamed the component to `RadioButton` (removed `Errorable`)
2. We're moving `formation-react` to a new repo: [`component-library`](https://github.com/department-of-veterans-affairs/component-library)
    - If you wanna make this change here in `formation-react`, that's fine; we'll port it over
    - Alternatively, you can make a new PR to `component-library` for this (I don't _think_ it's easy to move this PR to a new repo :thinking: )

File: packages/formation-react/src/components/RadioButtons/RadioButtons.jsx [hunk 2]
```diff
@@ -48,7 +47,16 @@ class RadioButtons extends React.Component {
     return null;
   }
 
-  handleChange(domEvent) {
+  handleChange(domEvent, optionLabel) {
+    // If a Google Analytics prop was passed in, fire the standardized event onChange
+    if (this.props.standardGAEvent && window.dataLayer) {
```
bkjohnson, line 52: Additionally, `standardGAEvent` needs to be documented in `propTypes`.

bkjohnson (changes requested):

bkjohnson:
Closing this PR - please migrate the changes to the [`component-library`](https://github.com/department-of-veterans-affairs/component-library) repo if they are still desired.

PR #557, opened 2020-12-10
Fix the pre-publish script

cvalarida:
## Description
Turns out `execSync` returns a buffer, which makes total sense. This buffer contained a newline character, which explained the dumb thing I was doing in the `console.log()`.

## Testing done
The latest `master` needed publishing, so I made the change there and tested by publishing. On this branch, I added a `process.exit(1)` at the end of the pre-publish script for safety and tried publishing again. I got the error I expected (wouldn't let me publish).
image

### DIFF
--- a/scripts/prepublish.js
+++ b/scripts/prepublish.js
@@ -6,14 +6,16 @@
 const { execSync } = require('child_process');
 const chalk = require('chalk');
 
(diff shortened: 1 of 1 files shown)


bkjohnson (approved):
:+1:

U-DON (approved):

PR #558, opened 2020-12-10
Build .js files

cvalarida:
## Description
Telephone.js was invalid because it was importing `./contacts`, but `contacts.js` wasn't being built by our build process because it was a `.js` file and not a `.jsx` file. This fixes that.

## Testing done
Built locally and verified `contacts.js` is built as expected.


bkjohnson (approved):
:shipit:

PR #562, opened 2020-12-18
Back to top

bkjohnson:
## Description

For issue: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/298

This adds a "Back to top" component that can be used in static pages.

Additionally, when it goes back to the top it tries to set focus on a `<body>` tag with a class of `merger`. This is based on the copied code, which is based on this comment: https://github.com/department-of-veterans-affairs/va.gov-team/issues/16737#issuecomment-739055927

Based on the following files from `vets-website`:

- https://github.com/department-of-veterans-affairs/vets-website/blob/5e8b75e201ea65bc651d17b6f474f09219cc1e1c/src/site/components/up_to_top_button.html
- https://github.com/department-of-veterans-affairs/vets-website/blob/58109bac9e82fb3195a4732da5066eb0484fff92/src/applications/static-pages/scroll-top-button.js
- https://github.com/department-of-veterans-affairs/vets-website/blob/8394a595db5ceb3c50a69c82b005377aa3fef82e/src/applications/static-pages/sass/modules/_scroll-top-button.scss

This hasn't been directly tested with the Medallia/Foresee `Feedback` button since I haven't been able to get either one to appear locally (they are loaded by an external script), but based on the behaviors I expect the conflict to be limited to mobile, and only when the user has scrolled down to the footer.

## Testing done

Storybook

## Screenshots

Here is an example of the button being flush with an accordion:

image

<details>
   <summary> Mobile view</summary>

Peek 2021-01-06 16-04

</details>

<details>
    <summary>Desktop view</summary>
   
Peek 2021-01-06 15-59

 
</details>

<details>
    <summary>Feedback button</summary>

    
Peek 2021-01-05 10-18

</details>

Notice that the `Feedback` button is anchored to the right side in the middle of the page. The BTT button is anchored at the right side, but on the bottom of the page. The only time these might conflict should be on mobile when a user scrolls down to the footer. The BTT button wouldn't go past the footer, but the Feedback button would appear to pass over top of it if the user scrolled to the very bottom of the page.

## Acceptance criteria
- [ ]

## Definition of done


CrystabelReiter (changes requested):
For the desktop component, can we reduce the padding between the arrow icon and "Back to top" label to only 8px?

packages/formation/sass/modules/_m-back-to-top.scss
```diff
@@ -0,0 +1,73 @@
+//=============================
+// Up To Top Button Styles
+//=============================
+#top-button-container {
```
cvalarida, line 4: How do you feel about `back-to-top-container` to avoid potential namespace collision.

packages/formation/sass/modules/_m-back-to-top.scss
```diff
@@ -0,0 +1,73 @@
...
+  bottom: 0;
+  pointer-events: none;
+  position: fixed;
+  width: 100%;
+
+  &.va-top-button-container-relative {
```
cvalarida, line 10: ~How and why is this used? (Should it be documented in Storybook?)~
Ahhhhh, I see. It's to toggle the "stickiness" when the footer is scrolled into view. Maybe make a comment to that effect here?

packages/formation/sass/modules/_m-back-to-top.scss
```diff
@@ -0,0 +1,73 @@
...
+    }
+    &.va-top-button-transition-in {
+      pointer-events: all;
+      opacity: 1;
+    }
+    @include media($medium-screen) {
+      padding-left: 17px;
+      padding-right: 19px;
+      width: auto;
+    }
```
cvalarida, line 73: Prolly should override border-radius here for the right corners so it's not pointy on that end and round on the other.

packages/formation/js/back-to-top.js
```diff
@@ -0,0 +1,63 @@
...
+  if (el) el.focus();
+
+  return window.scrollTo(0, 0);
+}
+
+function isScrolledIntoView(el) {
+  const elemTop = el.getBoundingClientRect().top;
+  // Only partially || completely visible elements return true
+  return elemTop >= 0 && elemTop <= window.innerHeight;
+}
```
cvalarida, line 17: Because this is triggered _every_ scroll event (which happens notoriously frequently) on _every_ page, have you done any performance benchmarking on this function? The rest of the event listener looks as fast as I can think to make it, but I'm not sure about `el.getBoundingClientRect()`.

How does it react when you use your dev tools to [throttle the CPU way down](https://umaar.com/dev-tips/88-cpu-throttling/)?

packages/formation/js/back-to-top.js
```diff
@@ -0,0 +1,62 @@
+/**
+ * This button is used for navigating the user back to the top of the page.
+ * It's to be used on long article pages.
+ */
+
+function navigateToTop() {
+  const el = document.querySelector('body.merger');
```
cvalarida, line 7: Do we need the `.merger`? That class is a vestige of the past (Vets.gov -> VA.gov web brand consolidation) and I'd love to see us kill it with fire at some point.

cvalarida (approved):
Had a couple of nit-picky comments, but overall, it looks pretty good. :+1:

packages/formation/sass/modules/_m-back-to-top.scss
bkjohnson: This is how it is supposed to be according to the documentation. From [zeplin](https://app.zeplin.io/project/5e29e7ff922d4d961b5bff4d/screen/5fdbda435599ca472d759b4c)

image

packages/formation/js/back-to-top.js
bkjohnson: Probably not, but in Storybook there _are_ multiple `<body>` tags. I'll play with this and see what works.

packages/formation/sass/modules/_m-back-to-top.scss
cvalarida: Ohhhhh, I seeeeeeee.

Is this what you intended, @CrystabelReiter?
image

It feels to me like either it should be hard against the scrollbar on the right—like it is in mobile—or a floating, rounded button.

CrystabelReiter: It looks a little funky here, but in Ryan's mocks it's supposed to sit flush on the left of the content 
Screen Shot 2021-01-06 at 3 13 39 PM

CrystabelReiter: @bkjohnson There still looks like there's a lot of space between the arrow icon and the button label. Can we reduce to 8px between the label and the icon?

packages/formation/js/back-to-top.js
bkjohnson: I haven't done any benchmarking. One thing that I was interested in using was [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API), but I would have had to add polyfills for IE and felt like it just wasn't worth it. However, I think that would improve performance.

packages/formation/js/back-to-top.js
bkjohnson: Looks like the `<iframe>` limits the scope appropriately, so the `<body>` ends up being the correct one.

packages/formation/sass/modules/_m-back-to-top.scss
bkjohnson: @cvalarida where did you get that screenshot? Right now the `0.5em` right margin evaluates to 7.5px in storybook, I'll see if I can get it to work with the 8px unit base we have set for va.gov.

image

CrystabelReiter (approved):

packages/formation/js/back-to-top.js
cvalarida: Okay, cool. It seemed pretty safe, but I wanted to verify. :+1:

packages/formation/sass/modules/_m-back-to-top.scss
cvalarida: Oh, I answered this in Slack, but for posterity: I took a screenshot of your gif. :grin:

packages/formation/js/back-to-top.js
ncksllvn: MDN recommends throttling using `window.requestAnimationFrame`, https://developer.mozilla.org/en-US/docs/Web/API/Document/scroll_event

PR #584, opened 2021-01-25
[Print view] Prepend website domain only when links start with "/"

ncksllvn:
## Description
Currently, in print preview, we append `https://va.gov${href}` during print preview, so that the printed page still tells users where the links go. This PR does not debate whether we should really be doing that or not -  it's just fixing some obvious issues. See [Slack discussion](https://dsva.slack.com/archives/C7NJWUA8K/p1611093271006000) with design system council.

The issues this fixes are -
- Prepend the domain onto the HREF if the HREF starts with a leading `/`
- Show the plain HREF if the HREF starts with `http`

https://github.com/department-of-veterans-affairs/va.gov-team/issues/18337
## Testing done
Observed `/coronavirus-veteran-frequently-asked-questions/` using the ["print" render mode](https://stackoverflow.com/questions/9540990/using-chromes-element-inspector-in-print-preview-mode)

##### Bug
image

##### Fixed
image

##### Bug
image

##### Fixed
image

## Acceptance criteria
- [ ] Obvious issues with prepending the domain during print view are fixed

## Definition of done

### DIFF
--- a/packages/formation/sass/modules/_m-print.scss
+++ b/packages/formation/sass/modules/_m-print.scss
@@ -20,12 +20,12 @@
   }
 
   /* Relative URLs - include the HREF with the domain prepended */
-  a:not([href^="#"]):not([href^="tel"]):not([href^="http"]):after {
+  a[href^="/"]:after {
     content: " (https://www.va.gov" attr(href) ")";
   }
 
   /* Absolute URLs - show the HREF without prepending the domain. */
-  a[href^="tel"]:after, a[href^="http"]:after {
+  a[href^="http"]:after {
     content: " (" attr(href) ")";
   }
 }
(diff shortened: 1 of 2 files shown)


packages/formation/sass/modules/_m-print.scss
```diff
@@ -8,19 +8,24 @@
...
     display: block;
   }
-  
-  /* Include href when printing */
-  a:not([href^="tel"]):after {
-    content: " (https://va.gov" attr(href) ")";
+
+  /* Relative URLs - include the HREF with the domain prepended */
+  a[href^="/"]:after {
+    content: " (https://www.va.gov" attr(href) ")";
```
ncksllvn, line 24: As shown in the diff, we previously omitted the `www`. This doesn't seem right to me it - it only saves four characters, so it's not as though we save a bunch of space on the page, and creates an unnecessary redirect. It also just doesn't look right to me to leave out the `www`.

cvalarida (approved):
Solid improvement! :+1: :shipit:

PR #585, opened 2021-01-26
Fix mobile side nav focus bug

erikphansen:
This fixes a bug in Safari where focus was not properly managed when opening the mobile side nav.

### DIFF
--- a/packages/formation/js/sidenav.js
+++ b/packages/formation/js/sidenav.js
@@ -95,6 +95,7 @@ class SideBarMenu {
     // set when the menu opened. This event loop hack fixes the issue and
     // ensures that the menu's close button gets focus when the sidenav menu
     // opens.
+    // https://github.com/department-of-veterans-affairs/va.gov-team/issues/18916
     setTimeout(() => {
       this.closeControl.focus();
     }, 0);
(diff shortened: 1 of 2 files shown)


packages/formation/js/sidenav.js
```diff
@@ -91,8 +91,13 @@ class SideBarMenu {
...
     this.addCloseMenuListener();
-    // main trigger, switch focus to close button
-    this.closeControl.focus();
+    // Switch focus to close button. In Safari the focus was not always getting
+    // set when the menu opened. This event loop hack fixes the issue and
+    // ensures that the menu's close button gets focus when the sidenav menu
+    // opens.
+    setTimeout(() => {
+      this.closeControl.focus();
+    }, 0);
```
bkjohnson, line 100: If this is a known issue, can you add a link to somewhere where it's documented?

bkjohnson (approved):
Weird.

packages/formation/js/sidenav.js
erikphansen: you mean a link to the github issue in a code comment?

erikphansen: I'll link to https://github.com/department-of-veterans-affairs/va.gov-team/issues/18916 which has a video showing focus not going to the close button like it should.

bkjohnson: That works!

PR #622, opened 2021-05-19
updated default build command for vets-website

timwright12:
## Description
We updated the default build command, this is the accompanying documentation.

### DIFF
--- a/packages/documentation/src/pages/getting-started/index.mdx
+++ b/packages/documentation/src/pages/getting-started/index.mdx
@@ -91,7 +91,7 @@ These instructions cover building and running VA.gov locally, including configur
    $ yarn install
    ```
 
-8a. Build `vets-website`. If you're only doing application work inside `vets-website` you can build which contains the scaffolding option by default:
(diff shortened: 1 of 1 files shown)


packages/documentation/src/pages/getting-started/index.mdx
```diff
@@ -91,9 +91,9 @@ These instructions cover building and running VA.gov locally, including configur
    $ yarn install
    ```
 
-8a. Build `vets-website`. If you're only doing application work inside `vets-website` you can build with the scaffold option:
+8a. Build `vets-website`. If you're only doing application work inside `vets-website` you can build which contains the scaffolding option by default:
```
oseasmoran73, line 94:
```suggestion
8a. Build `vets-website`. If you're only doing application work inside `vets-website` you can run the build command which contains the scaffolding option by default:
```

oseasmoran73 (approved):

packages/documentation/src/pages/getting-started/index.mdx
Cgfor3, line 94: nit - Phrasing seems a bit off with this change

Cgfor3 (approved):
Looks good!

PR #648, opened 2021-07-28
Fallback to window.pageYOffset for IE11

Rleahy22:
## Description
https://github.com/department-of-veterans-affairs/va.gov-team/issues/25103
I believe this is the result of `window.scrollY` being undefined in IE11: https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollY#browser_compatibility

## Testing done
unit testing

## Acceptance criteria
- [ ]

## Definition of done
- [ ] Changes have been tested in vets-website
- [ ] Changes have been tested in IE11, if applicable
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)
- [ ] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

PR #654, opened 2021-08-03
Run breakpointCheck even when footer is not present

Rleahy22:
## Description
https://github.com/department-of-veterans-affairs/va.gov-team/issues/25103

This is a follow up on https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/648 which ensured IE compatibility for the back to top component. The issue still persists and it appears to be a result of an unrelated check for the existence of the footer component before executing the breakpoint check on scroll. Currently we run checks for 2 separate elements before running the code that relies on the existence of either of them. This PR removes the too early check for `footer` and moves that check to where it's needed. See the video for the effect of both lines in this pr on IE11

https://user-images.githubusercontent.com/3144003/128720680-4ddc51fa-95b3-41c5-be64-523fb5e56649.mov

## Testing done
local via browserstack and unit

## Screenshots
Screen Shot 2021-08-03 at 11 53 23 AM

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/formation/js/back-to-top.js
+++ b/packages/formation/js/back-to-top.js
@@ -42,7 +42,7 @@ export function closure(button, buttonContainer, footer, buttonClasses) {
       hasHitBreakpoint = !hasHitBreakpoint;
     }
 
-    if (footerCheck() !== footerVisChanged) {
+    if (footer && footerCheck() !== footerVisChanged) {
       buttonContainer.classList.toggle(buttonClasses.containerRelative);
       footerVisChanged = !footerVisChanged;
     }
@@ -58,7 +58,6 @@ export default function setup() {
   // The current page likely does not contain a "Back to top" button in its layout.
 
   const footer = document.getElementById('footerNav');
-  if (!footer) return;
 
   const buttonClasses = {
     transitionIn: 'va-top-button-transition-in',
--- a/packages/formation/js/back-to-top.js
+++ b/packages/formation/js/back-to-top.js
@@ -20,7 +20,12 @@ function navigateToTop() {
 }
 
 function isScrolledIntoView(el) {
-  const elemTop = el.getBoundingClientRect().top;
+  const elemTop = el?.getBoundingClientRect().top;
+
+  if (!elemTop && elemTop !== 0) {
+    return false;
+  }
+
   // Only partially || completely visible elements return true
   return elemTop >= 0 && elemTop <= window.innerHeight;
 }
@@ -42,7 +47,7 @@ export function closure(button, buttonContainer, footer, buttonClasses) {
       hasHitBreakpoint = !hasHitBreakpoint;
     }
 
-    if (footer && footerCheck() !== footerVisChanged) {
+    if (footerCheck() !== footerVisChanged) {
       buttonContainer.classList.toggle(buttonClasses.containerRelative);
       footerVisChanged = !footerVisChanged;
     }
--- a/packages/formation/__tests__/js/back-to-top/back-to-top.test.js
+++ b/packages/formation/__tests__/js/back-to-top/back-to-top.test.js
@@ -82,6 +82,15 @@ describe('static page back to top widget', () => {
       .to.be.true;
   });
 
+  it('should toggle button transition class even when footer is not present', () => {
+    global.window.scrollY = 601;
+
+    closure(testButton, testButtonContainer, null, buttonClasses)();
+
+    expect(testButton.classList.toggle.calledWith(buttonClasses.transitionIn))
+      .to.be.true;
+  });
+
   it('should not toggle button container relative class when footer is out of view', () => {
     closure(
       testButton,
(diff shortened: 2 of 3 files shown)


kelsonic (approved):

Rleahy22:
Could I get a review from @department-of-veterans-affairs/vsp-design-system-fe? Thanks!

Rleahy22:
@bkjohnson @cvalarida could I get 👀 ?

packages/formation/js/back-to-top.js
```diff
@@ -58,7 +58,6 @@ export default function setup() {
   // The current page likely does not contain a "Back to top" button in its layout.
 
   const footer = document.getElementById('footerNav');
-  if (!footer) return;
```
bkjohnson, line 61: I'm a bit confused about how these changes solve the problem with the button not appearing in IE11. I see how not returning early will add the `scroll` handler, which will result in `breakpointCheck()` being called:


https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/c74674449c4a74018e5bf239e5069edc2fe75737/packages/formation/js/back-to-top.js#L39-L43

packages/formation/js/back-to-top.js
```diff
@@ -42,7 +42,7 @@ export function closure(button, buttonContainer, footer, buttonClasses) {
       hasHitBreakpoint = !hasHitBreakpoint;
     }
 
-    if (footerCheck() !== footerVisChanged) {
+    if (footer && footerCheck() !== footerVisChanged) {
```
bkjohnson, line 45: But it isn't clear why this is here.

bkjohnson (commented):
Can you explain how these changes allow the button to appear in IE11? I read the issue and saw an explanation for an earlier fix related to syntax, but I didn't see anything related to a footer.

packages/formation/js/back-to-top.js
Rleahy22: before this change we checked if `footer` existed before running either `breakpointCheck` or `footerCheck`. Since I removed the earlier check for `footer` I moved that check here, where it gets checked before the code that relies on it.

bkjohnson:
It's still unclear how this (presumably) solves the problem in the linked issue about the Back to Top button not displaying in IE11. What does a footer have to do with IE11?

The Platform has a [list of best practices for PRs](https://depo-platform-documentation.scrollhelp.site/developer-docs/Pull-request-best-practices.688259145.html#Pullrequestbestpractices-ClearlysummarizethechangeinthePRtitleanddescription):

> Explain how code changes are related to the pull request objectives.

Rleahy22:
@bkjohnson I appreciate your guidance on the description. I have updated it to explain why this code change is necessary. If you could please take another look, I would appreciate it.

packages/formation/js/back-to-top.js
bkjohnson: Without this addition, we just run `footerCheck()`, which runs `isScrolledIntoView(footer)`. It seems like that is returning truthy which is a problem. The error I'm seeing is `Unable to get property getBoundingClientRect`. Rather than avoid calling the function, what do you think about modifying the helper to return false if the element isn't there? Optional chaining may work:

```js
/**
 * Check if an element is scrolled into view.
 * Returns false if the element is null
 */
function isScrolledIntoView(el) {
  const elemTop = el?.getBoundingClientRect().top;
  // Only partially || completely visible elements return true
  return elemTop >= 0 && elemTop <= window.innerHeight;
}
```

(This is untested)

bkjohnson (commented):
At about 45 seconds into the video it looked like the page under test did not have a footer, and it is my understanding that all pages on va.gov have a footer. Is this fix for an issue that Veterans are experiencing, or is it for content editors?

packages/formation/js/back-to-top.js
Rleahy22: The transpiled version of this code is `el && el.getBoundingClientRect` which is functionally the same as `footer && footerCheck`, only checking for the existence of the element before even calling `footerCheck` avoids that unnecessary call entirely.

Rleahy22:
> At about 45 seconds into the video it looked like the page under test did not have a footer, and it is my understanding that all pages on va.gov have a footer. Is this fix for an issue that Veterans are experiencing, or is it for content editors?

I visited the same exact url in both Chrome and IE11. In Chrome the footer was there and the code executed as expected. In IE11 the footer was not there and the code did not execute. I am not sure why the footer is not available in IE11, honestly it is very hard for me to debug IE11 locally within MacOS. It does appear that we can solve the issue that Michelle reported with this change, and even then we currently have a logic gate preventing code from running that does not rely on that logic gate, so it seems like a reasonable change.

packages/formation/js/back-to-top.js
bkjohnson: Even if `footer && footerCheck()` stays in, I think it is better for us to return `false` when `footer` is `null` rather than throw an unexpected error.

bkjohnson (approved):
Don't forget to bump the version in `package.json` if you want this asap.

PR #663, opened 2021-08-11
Update dependencies

cvalarida:
## Description
This PR updates some woefully out-of-date dependencies. I did this so I could build the silly thing.

Previously, I was getting errors even just installing the dependencies. (This may have been because the repo is set up to build everything after installing the dependencies. I didn't look into _why_ it was doing what it was doing too closely.)

## Testing done
- Dependencies installed successfully
- Packages built successfully

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,12 +79,11 @@
     "karma-chrome-launcher": "^2.2.0",
     "karma-mocha": "^2.0.1",
     "karma-webpack": "^2.0.5",
-    "lerna": "^3.10.5",
+    "lerna": "^4.0.0",
     "mini-css-extract-plugin": "^0.5.0",
     "mkdirp": "^0.5.1",
     "mocha": "^4.0.1",
     "ncp": "^2.0.0",
-    "node-sass": "^4.8.3",
     "postcss-loader": "^3.0.0",
     "promisify-node": "^0.5.0",
     "react": "^16.7.0",
@@ -94,7 +93,8 @@
     "react-test-renderer": "^16.7.0",
     "recast": "^0.20.4",
     "resolve-url-loader": "^2.2.0",
-    "sass-loader": "^6.0.7",
+    "sass": "^1.37.5",
+    "sass-loader": "^12.1.0",
     "semver": "^7.3.2",
     "sinon": "^4.1.6",
     "style-loader": "^0.19.0",
@@ -102,7 +102,7 @@
     "url-loader": "^0.6.2",
     "uswds": "1.6.10",
     "vagov-eslint": "file:./packages/vagov-eslint",
-    "webpack": "4.29.6"
+    "webpack": "^5.50.0"
   },
   "dependencies": {
     "graphql": "^14.5.3"
--- a/packages/documentation/package.json
+++ b/packages/documentation/package.json
@@ -33,7 +33,6 @@
     "gatsby-transformer-sharp": "^2.9.0",
     "gh-pages": "^3.0.0",
     "node-github-graphql": "^0.2.7",
-    "node-sass": "^4.11.0",
     "prop-types": "^15.6.2",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -79,12 +79,11 @@
     "karma-chrome-launcher": "^2.2.0",
     "karma-mocha": "^2.0.1",
     "karma-webpack": "^2.0.5",
-    "lerna": "^3.10.5",
+    "lerna": "^4.0.0",
```
cvalarida, line 82: This wasn't strictly speaking necessary, but I was chasing down rabbits and...well, it's now at the latest version and everything still works as expected as far as I can tell.

package.json
```diff
@@ -79,12 +79,11 @@
...
     "mini-css-extract-plugin": "^0.5.0",
     "mkdirp": "^0.5.1",
     "mocha": "^4.0.1",
     "ncp": "^2.0.0",
-    "node-sass": "^4.8.3",
```
cvalarida, line 87: Not only is this woefully out of date, it's also deprecated entirely in favor of `sass`. This is the module that was giving me grief because it was calling `node-gyp@3.8.0` and that was throwing some errors in the `g++` compiler for...baffling reasons.

package.json
```diff
@@ -94,15 +93,16 @@
     "react-test-renderer": "^16.7.0",
     "recast": "^0.20.4",
     "resolve-url-loader": "^2.2.0",
-    "sass-loader": "^6.0.7",
+    "sass": "^1.37.5",
+    "sass-loader": "^12.1.0",
```
cvalarida, line 97: Updated to the latest version so I could use `sass` as the underlying sass compiler instead of `node-sass`. I combed through the release notes and didn't find any breaking changes that would affect us as far as I could tell.

package.json
```diff
@@ -94,15 +93,16 @@
...
     "semver": "^7.3.2",
     "sinon": "^4.1.6",
     "style-loader": "^0.19.0",
     "svg-url-loader": "^2.3.0",
     "url-loader": "^0.6.2",
     "uswds": "1.6.10",
     "vagov-eslint": "file:./packages/vagov-eslint",
-    "webpack": "4.29.6"
+    "webpack": "^5.50.0"
```
cvalarida, line 105: Updating `sass-loader` required a minimum `webpack` version of 5, so I updated this to the latest.

packages/documentation/package.json
```diff
@@ -33,23 +33,22 @@
...
-    "node-sass": "^4.11.0",
     "prop-types": "^15.6.2",
     "puppeteer": "^2.0.0",
     "react": "^16.6.3",
     "react-dom": "^16.6.3",
     "react-helmet": "^6.1.0",
     "react-live": "1.12.0",
     "uswds": "1.6.10",
-    "web-components": "https://github.com/department-of-veterans-affairs/component-library.git#wc-v0.5.0"
+    "web-components": "https://github.com/department-of-veterans-affairs/component-library.git#wc-v0.9.0"
```
cvalarida, line 43: Not strictly speaking necessary for what I was trying to do, but while I was here. :man_shrugging:

packages/documentation/package.json
```diff
@@ -59,7 +58,6 @@
   ],
   "license": "MIT",
   "scripts": {
-    "postinstall": "cd node_modules/web-components && yarn install --production && yarn build",
```
cvalarida, line 62: No longer needed with the updated `web-components` version; we moved this to that library's `prepare` script where it belongs.

package.json
theodur: Yeah, it doesn't look like any of the [breaking changes](https://github.com/lerna/lerna/blob/main/CHANGELOG.md#breaking-changes) should affect anything.

theodur (approved):

PR #684, opened 2021-09-20
Updated css to change the halo focus color

noahgelman:
## Description
Adds a new color and updates the focus color

https://app.zenhub.com/workspaces/vsp-design-system-5f8de67192551b0012ebb802/issues/department-of-veterans-affairs/vets-design-system-documentation/359

2nd related PR: https://github.com/department-of-veterans-affairs/component-library/pull/175

## Definition of done
- [ ] Not blue colored focused elements have an olive colored focus ring

### DIFF
--- a/packages/formation/package.json
+++ b/packages/formation/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "6.16.4",
+  "version": "6.16.5",
   "description": "The VA design system",
   "keywords": [
     "va",
--- a/packages/formation/sass/base/_b-mixins.scss
+++ b/packages/formation/sass/base/_b-mixins.scss
@@ -87,8 +87,13 @@
   }
 }
 
+@mixin focus-olive-outline($offset: 2) {
+  outline: $focus-olive-outline;
+  outline-offset: #{$offset}px;
+}
+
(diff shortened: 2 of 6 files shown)


bkjohnson (commented):
From the linked issue:

> The ultimate solution would be an inverted look a la GOV.UK but this is just to mitigate the current problematic yellow-on-light situation.

Does this mean that the olive solution is only temporary? I'm asking because I'm wondering how to implement this in our web components. Right now we have a [mixin](https://github.com/department-of-veterans-affairs/component-library/blob/web-components/src/mixins/focusable.css) that we import when we're defining a component that can grab focus.

noahgelman:
@bkjohnson Can you re-review? Would like to move to staging for additional review

bkjohnson (commented):
The change seems ok, but I'd like to hold off on merging & publishing a new release until a new PR for https://github.com/department-of-veterans-affairs/component-library/pull/175 is also ready to go. Publishing `formation` and `component-library` at the same time seems like a good plan so that the outlines will be the same regardless of the kind of component being used.

timwright12:
@bkjohnson Do you want to close this one out?

bkjohnson:
Closing, work will continue for https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/654

PR #699, opened 2021-09-30
Button focus

rjohnson2011:
## Description
Resolves https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/235 by making focus & hover states the same background shade for primary buttons and gives hover & focus a background shade for secondary buttons.

## Testing done
Reviewed in Storybook component (see screenshots below)

## Screenshots
Focus & Hover now have same background for Secondary Buttons.
image
Blue Default buttons now have same background color on hover and focus.
image

## Acceptance criteria
x Blue button has the same background color on hover and focus
x Blue button retains the yellow halo on focus
x Outlined secondary button has the same darker border and text on hover and focus
x Outlined secondary button retains the yellow halo on focus
x Create and/or modify unit tests for this component to include an aXe scan

### DIFF
--- a/packages/formation/sass/base/_b-mixins.scss
+++ b/packages/formation/sass/base/_b-mixins.scss
@@ -74,7 +74,7 @@
     padding: 0;
     border: 0;
     margin: -1px;
-    content: "this will open a new website";
+    content: 'this will open a new website';
   }
 }
 
@@ -102,9 +102,14 @@
   @include focus-gold-light-outline;
 }
 
+@mixin focus-dark-outline($offset: 2) {
+  outline: $color-primary-darker;
+  outline-offset: #{$offset}px;
+}
+
 @mixin color-transition {
   -webkit-transition-duration: 0.3s;
-    transition-duration: 0.3s;
+  transition-duration: 0.3s;
 
   -webkit-transition-timing-function: ease-in-out;
   transition-timing-function: ease-in-out;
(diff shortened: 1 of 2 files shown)


rjohnson2011:
@bkjohnson Can I get a review on this please?

packages/formation/sass/base/_b-mixins.scss
```diff
@@ -74,7 +74,7 @@
     padding: 0;
     border: 0;
     margin: -1px;
-    content: "this will open a new website";
+    content: 'this will open a new website';
```
bkjohnson, line 77: Did this file need to change?

packages/formation/sass/modules/_m-button.scss
```diff
@@ -3,33 +3,41 @@
...
+    &:focus,
+    &:hover {
+      background-color: $color-cool-blue-lightest;
+    }
   }
 }
 
+.usa-button:focus {
+  background: $color-primary-darker;
+}
```
bkjohnson, line 23: Should this be on `usa-button-primary` instead?

rjohnson2011: This effects the Primary Button section which is `.usa-button`
image

packages/formation/sass/base/_b-mixins.scss
rjohnson2011: No, got caught on my auto-linter. Reverted back to main branch, thanks!

bkjohnson (commented):
These blues are different - is that intentional?

image

bkjohnson (approved):

PR #707, opened 2021-11-15
Exclude BTT component from last-child uswds styling

bkjohnson:
## Description

Part of https://github.com/department-of-veterans-affairs/va.gov-team/issues/31200

`<va-back-to-top>` was designed to be the last node in the main content area. Our main content areas tend to have the `usa-width-three-fourths` CSS class on them. `uswds` has styling that [removes the bottom margin of the last child](https://github.com/uswds/uswds/blob/c4c9410d54b537c6112037a36ed289d8b7e8940a/src/stylesheets/elements/_typography.scss#L227-L231) in elements with that class. This results in `<va-back-to-top>` not looking like we expect (see screenshots)

This PR:

- Stops importing the `uswds` SCSS file responsible for removing the bottom margin on all last children
- Copies that SCSS file into our repo
- Modifies it to exclude `<va-back-to-top>` nodes from the `:last-child` styling.

This will also mean that https://github.com/department-of-veterans-affairs/content-build/pull/759 can just use `<va-back-to-top>` without adding an empty `<span>` afterwards to bypass being the last child.

## Testing done

`yarn link` on `formation` and a local content build. :eyes: 

### Bottom margin removed

image

### When last-child is not BTT

image

### When last-child _is_ BTT

image

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/formation/sass/base/_b-typography.scss
+++ b/packages/formation/sass/base/_b-typography.scss
@@ -1,0 +1,238 @@
+html {
+  font-family: $font-sans;
+  font-size: $em-base;
+}
+
+body {
+  font-size: $base-font-size;
+}
+
+p {
+  line-height: $base-line-height;
+  margin-bottom: 1em;
+  margin-top: 1em;
+}
+
+a {
+  color: $color-primary;
+  text-decoration: underline;
+
+  &:hover,
+  &:active {
+    color: $color-primary-darker;
+  }
+
+  &:focus {
+    @include focus;
+  }
+
+  &:visited {
+    color: $color-visited;
+  }
+}
+
+@mixin external-link($external-link, $external-link-hover) {
+  &::after {
+    background: url('#{$image-path}/#{$external-link}.png') no-repeat 0 0;
+    background: url('#{$image-path}/#{$external-link}.svg') no-repeat 0 0;
+    background-size: 100%;
+    content: '';
+    display: inline-block;
+    height: 0.65em;
+    margin-bottom: -1px;
+    margin-left: 4px;
+    width: 0.65em;
+  }
+
+  &:hover::after {
+    background-image: url('#{$image-path}/#{$external-link-hover}.png');
+    background-image: url('#{$image-path}/#{$external-link-hover}.svg');
+  }
+}
+
+// External link consider "effortless style approach":
+// [href^="http:"]:not([href*="my-domain.com"])
+// [href^="https:"]:not([href*="my-domain.com"])
+
+.usa-external_link {
+  @include external-link(external-link, external-link-hover);
+}
+
+.usa-external_link-alt {
+  @include external-link(external-link-alt, external-link-alt-hover);
+}
+
+h1,
+h2,
+h3,
+h4,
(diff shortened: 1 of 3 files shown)


packages/formation/sass/base/_b-typography.scss
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  > :first-child {
+    @include media($medium-screen) {
+      margin-top: 0;
+    }
+  }
+
+  &:last-child {
+    > :not(va-back-to-top):last-child {
```
bkjohnson, line 228: `:not(va-back-to-top)` is the only change made to this file.

packages/formation/sass/base/_b-typography.scss
```diff
@@ -0,0 +1,238 @@
...
+      margin-bottom: 0;
+    }
+  }
+
+  > :not(va-back-to-top):last-child {
```
bkjohnson, line 233: Here as well.

bkjohnson (commented):
This commit shows the styling change: https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/707/commits/a31307f84496a71401be0b4675953e4d5215c562

packages/formation/sass/base/_b-typography.scss
```diff
@@ -0,0 +1,238 @@
+html {
```
bkjohnson, line 1: This file was [copied from `uswds`](https://github.com/uswds/uswds/blob/v1.6.14/src/stylesheets/elements/_typography.scss), then the `va-back-to-top` changes were made.

k80bowman:
I'm curious, and I haven't dug into this myself so I may be way off base here, but is this an issue that could be solved by using greater specificity to override the USWDS styling rather than copying over the entire file and making changes?

bkjohnson:
We could probably use `!important` in the component's CSS to make sure it doesn't get overriden, but I haven't explored that because I really don't like using `!important`. On the formation side of things, I also don't like the idea of re-applying the margin after it has already been removed, because that seems like we're going against the encapsulation.

k80bowman:
I totally understand and I agree that `!important` isn't the way to go. I'm just not keen on using USWDS classes to refer to our own styling, if that makes sense. It makes the source of those styles unclear and will make it difficult to debug if issues arise in the future.

bkjohnson:
I'll see if I can find other options to explore.

bkjohnson:
We can talk further about the pros and cons of each of these during working hours, but here are the options I've come up with:

- `!important` on component
- Copy `uswds` file in formation w/ some changes (this PR)
- Use dummy elements in `content-build` so `<va-back-to-top>` is never the last or first child
- Add CSS with a higher specificity to `formation` or `component-library` globally to reapply margin
- Refactor `<va-back-to-top>` to add a wrapping div to Shadow DOM & apply margin to that
- Using multiple `:host` selectors to increase specificity

bkjohnson:
Closing in favor of https://github.com/department-of-veterans-affairs/component-library/pull/223

PR #727, opened 2022-01-20
Remove unused Webpack loaders

theodur:
## Description
There are a few Webpack loaders that aren't being used in the repo that we can remove.

## Testing done
Tested locally and in CI

## Definition of done


package.json
```diff
@@ -71,6 +71,7 @@
     "css-loader": "^0.28.7",
     "enzyme": "^3.1.0",
     "enzyme-adapter-react-16": "^1.7.1",
+    "file-loader": "4.0.0",
```
theodur, line 74: I have to manually add this because it's a lodaer that we use, but was never added here because of one of the loaders being removed required it as a dependency

alexandec (approved):
Getting rid of loaders, the best way to improve webpack 😂

PR #806, opened 2022-04-13
39750 Add Calc to Math Variables

pdavies88:
## Description
Add calc around math variables

## Closes
https://github.com/department-of-veterans-affairs/va.gov-team/issues/39750

## Acceptance criteria
- [ ] Add in calc

## Definition of done
- [X] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)

### DIFF
--- a/packages/documentation/yarn.lock
+++ b/packages/documentation/yarn.lock
@@ -10996,9 +10996,9 @@ moment-mini@^2.24.0:
   integrity sha512-9ARkWHBs+6YJIvrIp0Ik5tyTTtP9PoV0Ssu2Ocq5y9v8+NOOpWiRshAp8c4rZVWTOe+157on/5G+zj5pwIQFEQ==
 
 moment@^2.23.0, moment@^2.27.0:
-  version "2.29.1"
-  resolved "https://registry.yarnpkg.com/moment/-/moment-2.29.1.tgz#b2be769fa31940be9eeea6469c075e35006fa3d3"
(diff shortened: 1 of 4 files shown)


pdavies88:
@andresriveratoro @bkjohnson @k80bowman do you think we should publish a new version of Formation with this change?

k80bowman:
We'll need to in order to use this on our documentation site, correct? We'll want to spot check uses in vets-website and content-build before upgrading the dependency there, though.

k80bowman (approved):

pdavies88:
@k80bowman yes we would need to change the dependency to the new version for the documentation repo for this change to take effect. I will also test on vets-website and include a screen shot of the results

bkjohnson (commented):
Yeah, a patch version would be good to fix the sass warning in the documentation site.

bkjohnson (approved):

pdavies88:
@k80bowman @humancompanion-usds @bkjohnson this updates causes a failure on Vets Website as it forces the variable of $multiple to be locked and not converted to a pixel value.

Screenshot here of before the change:
Screen Shot 2022-04-14 at 9 13 17 AM

Screenshot after the change showing the failure:
Screen Shot 2022-04-14 at 9 23 15 AM

Since we use spacers for all of the `.vads-u-padding--` and `.vads-u-margin--` we probably should just leave as is and instead suppress the error message in the documentation site.

pdavies88:
Issue solved per Katy's comment: https://github.com/department-of-veterans-affairs/va.gov-team/issues/39750#issuecomment-1099196119

PR #813, opened 2022-04-29
adding vagov-platform package

timwright12:
## Description

Adding a package for vagov-platform with a shared utility for security audits on dependencies. We're using this in 3 places right now, so I wanted to publish it as a shared package to reduce redundancy in the system

## Testing done

Testing with vets-website and linked npm package


packages/vagov-platform/security/dependency-check/dependency-check.js
```diff
@@ -0,0 +1,106 @@
...
+const filepath = argv.filepath ? argv.filepath : './package.json';
+const absPath = resolve(process.cwd(), filepath);
+// eslint-disable-next-line import/no-dynamic-require
+const packageJSON = require(absPath);
+const exceptionSet = new Set([
+  'https://npmjs.com/advisories/996',
+  'https://npmjs.com/advisories/1488',
+  'https://github.com/advisories/GHSA-r683-j2x4-v87g',
+  'https://github.com/advisories/GHSA-8hfj-j24r-96c4',
+]);
```
theodur, line 15: Since we're turning this into a package, would it make sense to have the `exceptionSet` come from the repositories using this, rather than hard-coding it here? Not all the advisories in the exception set will apply to every repo using this.

I'm also thinking of times when we want to temporarily add an advisory to exception set to unblock CI while we're working on a resolution. It would be easier to set the advisories in the dependent repositories instead of having to publish an update to this package. 

Maybe we can have a config file with a list of advisories to ignore in the dependent repositories, where this script can pull from. That could be an option for the CLI, or be in a default place where this script always looks for the exception set.

timwright12: I think that's a good idea, but we also don't want to let repos/teams just skip critical security alerts like they do for linting right now. I don't mind blocking CI since we do that right now anyway.

theodur: Yeah that's a potential caveat, depending on how we set it up. Another concern of hard-coding the exception set here is when we want to remove old advisories from it. We won't have any way of knowing that removing an advisory from the set won't break CI for one of the repos using this package. That could cause the exception set to grow over time, creating tech debt. It might be worth making a ticket to handle/improve the exception set functionality for the security audit.

packages/vagov-platform/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+### yarn
+
+```bash
+yarn add @department-of-veterans-affairs/vagov-platform
+```
+
+## Dependency Security Checks
+
+Shared command to run secrity checks on non-dev dependencies.
```
theodur, line 19:
```suggestion
Shared command to run security checks on non-dev dependencies.
```

theodur (approved):

PR #830, opened 2022-06-08
Deprecate green button

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/653

## Testing done
Screen Shot 2022-06-08 at 10 02 13

## Acceptance criteria
- [x] ESLint rule to deprecate `va-button-primary` utility class exists

## Definition of done

### DIFF
--- a/packages/eslint-plugin/lib/config/recommended.js
+++ b/packages/eslint-plugin/lib/config/recommended.js
@@ -182,5 +182,6 @@ module.exports = {
     '@department-of-veterans-affairs/correct-apostrophe': 1,
     '@department-of-veterans-affairs/cypress-viewport-deprecated': 1,
     '@department-of-veterans-affairs/prefer-web-component-library': 1,
+    '@department-of-veterans-affairs/design-system': 1,
   },
 };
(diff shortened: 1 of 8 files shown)


k80bowman (approved):

it-harrison (approved):

packages/eslint-plugin/lib/config/recommended.js
```diff
@@ -182,5 +182,6 @@ module.exports = {
     '@department-of-veterans-affairs/correct-apostrophe': 1,
     '@department-of-veterans-affairs/cypress-viewport-deprecated': 1,
     '@department-of-veterans-affairs/prefer-web-component-library': 1,
+    '@department-of-veterans-affairs/design-system': 1,
```
bkjohnson, line 185: Can we make the rule name a bit more specific? Maybe something like `deprecated-classes`, or even very specific like `deprecate-primary-button`.

theodur (commented):
Don't forget to bump the version in the `package.json` 🙂

packages/eslint-plugin/lib/config/recommended.js
andresriveratoro: The intent here was to create a more generic rule name that we could reuse for non web component rules, but I can change it if that's not what we want.

bkjohnson: `deprecated-classes` sounds like it could be re-used for non web component rules unless I'm misunderstanding things.

it-harrison (approved):

PR #844, opened 2022-06-29
[41668] Add function for getting CI workflow runs

ndsprinkle:
## Description
This PR adds a function that uses octokit to GET the latest CI workflow runs for a provided repository. The function maps the response to a single object that consists of the commit SHAs as keys and the workflow run status as values.

## Acceptance criteria
- [ ] Function creating for fetching latest 30 CI workflow runs
- [ ] Function maps data to an object of commit SHA keys and workflow run status values

### DIFF
--- a/packages/documentation/src/components/dashboard/DashboardDataFetch.jsx
+++ b/packages/documentation/src/components/dashboard/DashboardDataFetch.jsx
@@ -1,3 +1,34 @@
+import { Octokit } from '@octokit/rest';
+
+function getWorkflowRunsAsCommitStatusObject(repo) {
+  const octokit = new Octokit();
+
+  const params = {
+    owner: 'department-of-veterans-affairs',
+    repo,
+    workflow_id: 'continuous-integration.yml',
+    branch: 'main',
+    per_page: 30,
+    page: 1,
+  };
+
+  return octokit.rest.actions.listWorkflowRuns(params).then(response => {
+    if (response.status !== 200) {
+      throw new Error(`Response ${response.status} from ${response.url}.`);
+    }
+    return response.data;
+  }).then(({ workflow_runs }) => {
+    if (workflow_runs.length === 0) {
+      throw new Error('No workflows found.');
+    }
+
+    return workflow_runs.reduce((map, obj) => {
+      map[obj["head_sha"]] = obj["status"];
+      return map;
+    }, {});
+  });
+}
+
 export async function DeployStatusDataFetch(repo) {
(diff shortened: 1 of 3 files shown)


packages/documentation/src/components/dashboard/DashboardDataFetch.jsx
```diff
@@ -1,3 +1,34 @@
...
+      throw new Error(`Response ${response.status} from ${response.url}.`);
+    }
+    return response.data;
+  }).then(({ workflow_runs }) => {
+    if (workflow_runs.length === 0) {
+      throw new Error('No workflows found.');
+    }
+
+    return workflow_runs.reduce((map, obj) => {
+      map[obj["head_sha"]] = obj["status"];
```
alexandec, line 26: I think we actually want `conclusion` instead of `status`. If status is `completed`, that means the workflow has finished running, but then the conclusion could be either `success` or `failure`. I find their choice of terminology a bit strange, but that's my understanding

packages/documentation/src/components/dashboard/DashboardDataFetch.jsx
```diff
@@ -1,3 +1,34 @@
...
+
+function getWorkflowRunsAsCommitStatusObject(repo) {
+  const octokit = new Octokit();
+
+  const params = {
+    owner: 'department-of-veterans-affairs',
+    repo,
+    workflow_id: 'continuous-integration.yml',
+    branch: 'main',
+    per_page: 30,
```
alexandec, line 11: One thing I'm not 100% sure of: if a workflow is re-run for a given commit, will that count as multiple runs here? Or does it only return the final run per commit (30 commits is 30 API results)?

ndsprinkle: I want to say it's the final run because I think it keeps the same `id` even when it's re-run, but I am not 100% sure on that. I had to deal with that situation back when I worked on the CI GHA.

ndsprinkle: Yeah I just manually re-ran my CI workflow for this PR and it kept its `id` and stayed in the same position in the response: https://api.github.com/repos/department-of-veterans-affairs/veteran-facing-services-tools/actions/workflows/continuous-integration.yml/runs

alexandec: Okay great, we should be good with 30 per page then. Thanks for checking!

alexandec (approved):

PR #863, opened 2022-08-23
Add ESLint rules for telephone numbers

bkjohnson:
## Description

Part of https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/864

This adds two new custom ESLint rules:

### prefer-telephone-component

This is an autofixable ESLint rule which will trigger on the pattern `<a href="tel:..."`. When run with `--fix`, it will replace that with `<va-telephone>`. It will also:

- replace the `href` attribute with `contact
- strip all non-digits out of the contact number
- remove child content since the component doesn't allow it
- remove the `aria-label` if present since the component will add its own to the link

### telephone-contact-3-or-10-digits

This is a rule added as part of this PR to try to prevent the first rule from applying a fix that results in the `<va-telephone>` component having a bad `contact` prop, i.e. one that doesn't have 3 or 10 digits. It might flag on code that results from the autofixable rule above, but also on [code like this](https://github.com/department-of-veterans-affairs/vets-website/blob/787b2486cf316ec1c5b4535866ee2d732e08bf26/src/applications/financial-status-report/components/AvailableDebts.jsx#L106).

This rule will try to remove any non-digit characters (i.e. hyphens) from the `contact` prop that it can, since they aren't digits and extend the length beyond 10.

This specific constraint is mentioned on [Storybook](https://design.va.gov/storybook/?path=/docs/components-va-telephone--default):

Screenshot of the telephone component's storybook page with the contact prop's description highlighted. It reads "3 or 10 digit string representing the contact number"

As well as on the [component's guidance page](https://design.va.gov/components/telephone#code-usage):

Screenshot of the component's page on design.va.gov. Under the "How to use va-telephone" heading the bullet point "Add a 3 or 10 digit phone number to the component to have it formatted correctly for usage in a page." is highlighted. Below that under "Code usage", in the properties table the description for `contact` is once again highlighted.

## Testing done

- Local unit testing with ESLint's `RuleTester`
- Testing in `vets-website`

## Screenshots

Local testing in `vets-website` using:

```diff
diff --git a/package.json b/package.json
index b8e8ea8e0c..217bac41b7 100644
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "@babel/traverse": "^7.17.3",
     "@bahmutov/cypress-esbuild-preprocessor": "^2.1.2",
     "@cypress/code-coverage": "^3.10.0",
-    "@department-of-veterans-affairs/eslint-plugin": "^1.2.2",
+    "@department-of-veterans-affairs/eslint-plugin": "link:../veteran-facing-services-tools/packages/eslint-plugin",
     "@department-of-veterans-affairs/generator-vets-website": "^3.8.0",
     "@octokit/rest": "^18.11.0",
     "@pact-foundation/pact": "^9.16.4",
```

A screenshot of vim with a file opened in the vets-website repo. There is an anchor link with a tel href, with an ESLint warning saying "The va-telephone web component should be used to display phone numbers".

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/eslint-plugin/lib/rules/prefer-telephone-component.js
+++ b/packages/eslint-plugin/lib/rules/prefer-telephone-component.js
@@ -1,4 +1,3 @@
-// const { elementType, getProp, getPropValue } from 'jsx-ast-utils';
 const jsxAstUtils = require('jsx-ast-utils');
 
 const { elementType, getProp, getPropValue } = jsxAstUtils;
@@ -10,7 +9,7 @@ module.exports = {
   meta: {
     docs: {
       description:
-        'Web Components are preferred over deprecated React component library components',
+        'The telephone Web Component is preferred over anchor links with tel hrefs',
       category: 'Best Practices',
       recommended: false,
     },
@@ -50,7 +49,7 @@ module.exports = {
                 ),
                 // The component will be self-closing since it accepts no child content
                 fixer.remove(node.closingElement),
-              ].filter(i => !!i);
+              ];
             },
           });
         }
--- a/packages/eslint-plugin/lib/rules/prefer-telephone-component.js
+++ b/packages/eslint-plugin/lib/rules/prefer-telephone-component.js
@@ -29,17 +29,14 @@ module.exports = {
 
         if (hrefValue.startsWith('tel:')) {
           const contact = hrefValue.substring('tel:'.length);
-          console.log('CONTACT', contact);
 
           context.report({
             node,
             message: MESSAGE,
             fix: fixer => {
-              // Replace the node name
-              // and remove the `pattern` prop if it's there
               return [
                 fixer.replaceText(node.name, 'va-telephone'),
-                fixer.replaceText(hrefProp.name, 'contact'),
+                fixer.replaceText(hrefProp, `contact="${contact}"`),
               ].filter(i => !!i);
             },
           });
--- a/packages/eslint-plugin/lib/rules/prefer-telephone-component.js
+++ b/packages/eslint-plugin/lib/rules/prefer-telephone-component.js
@@ -30,7 +30,9 @@ module.exports = {
         const hrefValue = getPropValue(hrefProp);
 
         if (hrefValue.startsWith('tel:')) {
-          const contact = hrefValue.substring('tel:'.length);
+          const contact = hrefValue
+            .substring('tel:'.length)
+            .replace(/[^\d]/g, '');
 
           context.report({
             node,
--- a/packages/eslint-plugin/lib/rules/prefer-telephone-component.js
+++ b/packages/eslint-plugin/lib/rules/prefer-telephone-component.js
@@ -39,11 +39,14 @@ module.exports = {
               return [
                 fixer.replaceText(anchorNode.name, 'va-telephone'),
                 fixer.replaceText(hrefProp, `contact="${contact}"`),
+                // Remove the children
                 ...node.children.map(c => fixer.remove(c)),
+                // Add the self-closing slash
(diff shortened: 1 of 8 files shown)


packages/eslint-plugin/lib/rules/prefer-telephone-component.js
```diff
@@ -0,0 +1,56 @@
...
+
+        const anchorNode = node.openingElement;
+
+        const hrefProp = getProp(anchorNode.attributes, 'href');
+        const hrefValue = getLiteralPropValue(hrefProp);
+
+        if (hrefValue?.startsWith('tel:')) {
+          const contact = hrefValue
+            .substring('tel:'.length)
+            .replace(/[^\d]/g, '');
```
bkjohnson, line 28: I decided against having this rule do any kind of length checking here. The [ESLint docs on applying fixes](https://eslint.org/docs/latest/developer-guide/working-with-rules#applying-fixes) say under best practices:

> Make fixes as small as possible. Fixes that are unnecessarily large could conflict with other fixes, and prevent them from being applied.

I didn't want to have a single rule rewrite code from an `<a>` to a `<va-telephone>` _and_ potentially truncate a phone number by removing some digits. I also didn't want to have this rule ignore `tel` links, so I wrote the second rule to take care of notifying developers when the `contact` prop doesn't have the right number of digits.

jamigibbs:
This will still complain for international numbers. Are we ok with that being an exception? Example at `src/applications/toe-benefits/components/underReview.jsx`

Screen Shot 2022-08-25 at 12 40 32 PM

jamigibbs:
I'm not sure if this is possible but will it allow us to control the spacing output? For example, prettier will complain that a space is needed before the end-slash and also sometimes there is a double space before `contact`:

Screen Shot 2022-08-25 at 12 43 58 PM

Screen Shot 2022-08-25 at 12 46 23 PM

bkjohnson:
I'm ok with that. That number also has two leading 0s, and in our documentation we say to use the `international` prop for international numbers. 

Screenshot of the telephone page on storybook with the "internation" prop highlighted in the args table. The description reads "Indicates if this is a number meant to be called from outside the US. Prepends a "+1" to the formatted number."

I was trying to avoid having this work creep into trying to detect if a number _should_ be marked as `international` or not since the original goal was to get `<a href="tel"` links to be a `<va-telephone>`.

Do you think I should add a third rule which looks for 11 digit contact numbers beginning with `1`, or 13 digit contact numbers beginning with `001` and removes those and adds an `international` prefix?

jamigibbs:
> Do you think I should add a third rule which looks for 11 digit contact numbers beginning with 1, or 13 digit contact numbers beginning with 001 and removes those and adds an international prop?

I can definitely see a bit of scope creep there and I'm fine with keeping this just for domestic numbers. But the warning message is confusing when you read it next to an international number. It makes it seem like there's a bug in the linter. 

Would things become too verbose if the warning text is just updated to make mention of internationl numbers being an exception or to use the international prop? Or even just "The <va-telephone> contact prop should only contain digits and be 3 or 10 digits long _for domestic numbers_."

bkjohnson:
What do you think about this for a message?

> The <va-telephone> contact prop should only contain digits and be 3 or 10 digits long. Consider the international prop if you want the number to begin with a 1.

jamigibbs:
> What do you think about this for a message?
> 
> > The  contact prop should only contain digits and be 3 or 10 digits long. Consider the international prop if you want the number to begin with a 1.

That's much clearer to me! 👍

jamigibbs (approved):

theodur (approved):

PR #868, opened 2022-09-09
File input web component migration rule

jamigibbs:
## Description
Create an ESLint rule for the FileInput component. 

Closes: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1048

## Testing done
VS Code

## Screenshots

Screen Shot 2022-09-12 at 3 43 38 PM

## Acceptance criteria
- [ ] Create an ESLint rule (warning) for the FileInput React component
- [ ] Notify teams using this component and encourage them to migrate to the web component version

## Definition of done

### DIFF
--- a/packages/eslint-plugin/lib/rules/prefer-web-component-library.js
+++ b/packages/eslint-plugin/lib/rules/prefer-web-component-library.js
@@ -369,6 +369,48 @@ const ombInfoTransformer = (context, node) => {
   });
 };
 
+const fileInputTransformer = (context, node) => {
+  const openingTagNode = node.openingElement.name;
+  const closingTagNode = node.closingElement?.name;
+  const buttonTextNode = getPropNode(node, 'buttonText');
+  const errorMessageNode = getPropNode(node, 'errorMessage');
+  const additionalClassNode = getPropNode(node, 'additionalClass');
+  const additionalErrorClassNode = getPropNode(node, 'additionalErrorClass');
+  const multipleNode = getPropNode(node, 'multiple');
+
+  context.report({
+    node,
+    message: MESSAGE  +
+    `\nBEWARE: multiple, additionalClass, and additionalErrorClass properties have been removed.` + 
+    `\nSee the File Input design system guidance: https://design.va.gov/components/form/file-input`,
+    data: {
+      reactComponent: openingTagNode.name,
+      webComponent: 'va-file-input',
+    },
+    suggest: [
+      {
+        desc: 'Migrate component',
+        fix: fixer => {
+          return [
+            // Rename component tags
+            fixer.replaceText(openingTagNode, 'va-file-input'),
+            closingTagNode && fixer.replaceText(closingTagNode, 'va-file-input'),
+
+            // Rename props if they exist
+            buttonTextNode && fixer.replaceText(buttonTextNode.name, 'button-text'),
+            errorMessageNode && fixer.replaceText(errorMessageNode.name, 'error'),
+
+            // Remove props that are no longer available
+            additionalClassNode && fixer.remove(additionalClassNode),
+            additionalErrorClassNode && fixer.remove(additionalErrorClassNode),
+            multipleNode && fixer.remove(multipleNode),
+          ].filter(i => !!i);
+        }
+      }
+    ]
+  })
+}
+
 /**
  * Stores the result of a check that determines if a component is part of
  * the Design System component-library.
@@ -450,6 +492,9 @@ module.exports = {
         switch (componentName) {
           case 'Breadcrumbs':
             breadcrumbsTransformer(context, node);
+            break;
+          case 'FileInput':
+            fileInputTransformer(context, node);
             break;
           case 'Modal':
             modalTransformer(context, node);
--- a/packages/eslint-plugin/lib/rules/prefer-web-component-library.js
+++ b/packages/eslint-plugin/lib/rules/prefer-web-component-library.js
@@ -377,28 +377,30 @@ const fileInputTransformer = (context, node) => {
   const additionalClassNode = getPropNode(node, 'additionalClass');
   const additionalErrorClassNode = getPropNode(node, 'additionalErrorClass');
   const multipleNode = getPropNode(node, 'multiple');
+  const changeHandler = getPropNode(node, 'onChange');
 
   context.report({
     node,
     message: MESSAGE  +
-    `\nBEWARE: multiple, additionalClass, and additionalErrorClass properties have been removed.` + 
+    `\nBEWARE: The multiple, additionalClass, and additionalErrorClass properties are no longer available.` + 
     `\nSee the File Input design system guidance: https://design.va.gov/components/form/file-input`,
     data: {
       reactComponent: openingTagNode.name,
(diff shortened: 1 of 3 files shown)


it-harrison (approved):

packages/eslint-plugin/lib/rules/prefer-web-component-library.js
```diff
@@ -369,6 +369,50 @@ const ombInfoTransformer = (context, node) => {
...
+        fix: fixer => {
+          return [
+            // Rename component tags (Bindings)
+            fixer.replaceText(openingTagNode, 'VaFileInput'),
+            closingTagNode && fixer.replaceText(closingTagNode, 'VaFileInput'),
+
+            // Rename props if they exist
+            buttonTextNode && fixer.replaceText(buttonTextNode.name, 'button-text'),
+            errorMessageNode && fixer.replaceText(errorMessageNode.name, 'error'),
+            changeHandler && fixer.replaceText(changeHandler.name, 'onVaChange'),
```
bkjohnson, line 403: The React component's `onChange` prop [passes a `FileList` object as the parameter](https://github.com/department-of-veterans-affairs/component-library/blob/788abc85f56ea02704b9f66fbb907924434eb295/packages/react-components/src/components/FileInput/FileInput.jsx#L17). For the web component, the `onVaChange` handler will receive an Event as its argument. I think doing this would put the component into a broken state.

packages/eslint-plugin/lib/rules/prefer-web-component-library.js
```diff
@@ -369,6 +369,50 @@ const ombInfoTransformer = (context, node) => {
...
+      reactComponent: openingTagNode.name,
+      webComponent: 'VaFileInput',
+    },
+    suggest: [
+      {
+        desc: 'Migrate component',
+        fix: fixer => {
+          return [
+            // Rename component tags (Bindings)
+            fixer.replaceText(openingTagNode, 'VaFileInput'),
```
bkjohnson, line 397: I think we would need to add an import for the React binding.

packages/eslint-plugin/lib/rules/prefer-web-component-library.js
```diff
@@ -369,6 +369,50 @@ const ombInfoTransformer = (context, node) => {
...
+            closingTagNode && fixer.replaceText(closingTagNode, 'VaFileInput'),
```
bkjohnson, line 398: It looks like `FileInput` doesn't have a `children` prop and is always used as a self-closing tag, so we might not need this.

bkjohnson (commented):
If we can't figure out a good way to make the `onChange`/`onVaChange` suggestion not break things, maybe the best thing to do is just close this PR and [configure the `deprecate/import` rule to trigger on `FileInput` as well](https://github.com/department-of-veterans-affairs/vets-website/blob/d3ae3b2ee933692e95fbd31885214071e0bcfe04/.eslintrc.js#L35-L68).

packages/eslint-plugin/lib/rules/prefer-web-component-library.js
jamigibbs: I'm not entirely sure what you mean by figuring out a good way to make `onChange/onVaChange` suggestion not break things.

The `onVaChange` event's payload is also accessed as `event.detail.files` and is a passthrough for the FileList it receives from the input: https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/components/va-file-input/va-file-input.tsx#L82

bkjohnson: I know that the payload can still be accessed, but the structure is different. The rule would offer a suggestion [on this bit of code](https://github.com/department-of-veterans-affairs/vets-website/blob/258384edbec8b96d18b4b500f0bf345a3b23c1ae/src/applications/financial-status-report/components/FileUploader.jsx#L84-L90), which has [this function as its `onChange` handler](https://github.com/department-of-veterans-affairs/vets-website/blob/258384edbec8b96d18b4b500f0bf345a3b23c1ae/src/applications/financial-status-report/components/FileUploader.jsx#L30). That function is expecting a [`FileList`](https://developer.mozilla.org/en-US/docs/Web/API/FileList) as its argument, because that's what the React component passes when it calls the function.

With the web component, `onVaChange` isn't just a function as a prop, it's an event handler, and its argument will probably be our custom `vaChange` event resembling an [`InputEvent`](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent).

Here's a more concrete example that I've collapsed so that the text below doesn't get shoved too far down.

<details>

<summary>Example with diff and screenshots</summary>

I made some local changes to log the argument that each function receives so we can see them.

```diff
diff --git a/packages/storybook/stories/FileInput.stories.jsx b/packages/storybook/stories/FileInput.stories.jsx
index 1b3ee04..a679ff0 100644
--- a/packages/storybook/stories/FileInput.stories.jsx
+++ b/packages/storybook/stories/FileInput.stories.jsx
@@ -22,7 +22,8 @@ const Template = args => (
 
 const defaultArgs = {
   buttonText: 'Upload some files',
-  onChange: () => {
+  onChange: onChangeArg => {
+    console.log('argument in React component:', onChangeArg);
     alert('uploaded file');
   },
 };
diff --git a/packages/storybook/stories/va-file-input.stories.jsx b/packages/storybook/stories/va-file-input.stories.jsx
index 6e0d60c..856dbd8 100644
--- a/packages/storybook/stories/va-file-input.stories.jsx
+++ b/packages/storybook/stories/va-file-input.stories.jsx
@@ -54,7 +55,12 @@ const Template = ({
 };
 
 export const Default = Template.bind(null);
-Default.args = { ...defaultArgs };
+Default.args = {
+  ...defaultArgs,
+  vaChange: onVaChangeArg => {
+    console.log('argument in web component:', onVaChangeArg);
+  },
+};
 
 Default.argTypes = propStructure(fileInputDocs);
 
```

### React component

Screenshot of the console open on a local storybook instance on the React file input component. A file was uploaded and then a console message was logged with a `FileList` object as the argument for the `onChange` function

### Web component

Screenshot of the console open on the same local storybook instance but the page is the File input web component. A file was uploaded and a console message appeared with a completely different object as the argument to the `vaChange` handler

</details>

So what I'm saying is, this ESLint suggestion would put things into a state that would break at runtime. [`uploadedFile[0]`](https://github.com/department-of-veterans-affairs/vets-website/blob/258384edbec8b96d18b4b500f0bf345a3b23c1ae/src/applications/financial-status-report/components/FileUploader.jsx#L31) won't exist in the web component version like it does in the React component because `uploadedFile` won't be indexable like that. Unfortunately we have to accept that developers don't always test changes, especially ones which come from an ESLint rule. It is just a _suggestion_ and not an autofix, but I do think that we shouldn't offer an automated suggestion like that if we know that it would be a breaking change. If we can rewrite the rule so it takes the changes to the function signature into account, great. But if that's impossible or infeasible I think we should try a simpler approach like the ESLint rule for deprecating the import.

jamigibbs:
Closing this PR in favor of a deprecate/import rule: https://github.com/department-of-veterans-affairs/vets-website/pull/22216

PR #900, opened 2022-12-13
Add ESLint rule for new CSS Classes

jhonnyoddball:
## Description

Part of https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1343

This adds two new custom ESLint rules:

### use-new-utility-classes

This Eslint rule will help VFS teams to easily update to the new CSS Class library based on USWDS v3. 

This is a recommendable ESLint rule which will trigger on the pattern `vads-u`. 

css rule eslint

## Testing done
- Testing in `vets-website`

## Screenshots
Local testing in `vets-website` using:

Screen Shot 2022-12-13 at 4 10 56 PM

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/eslint-plugin/lib/config/recommended.js
+++ b/packages/eslint-plugin/lib/config/recommended.js
@@ -185,6 +185,7 @@ module.exports = {
     '@department-of-veterans-affairs/prefer-telephone-component': 1,
     '@department-of-veterans-affairs/telephone-contact-digits': 1,
     '@department-of-veterans-affairs/deprecated-classes': 1,
+    '@department-of-veterans-affairs/use-new-utility-classes': 1,
     '@department-of-veterans-affairs/use-workspace-imports': 1,
   },
 };
--- a/packages/eslint-plugin/lib/config/recommended.js
+++ b/packages/eslint-plugin/lib/config/recommended.js
@@ -185,7 +185,6 @@ module.exports = {
     '@department-of-veterans-affairs/prefer-telephone-component': 1,
     '@department-of-veterans-affairs/telephone-contact-digits': 1,
     '@department-of-veterans-affairs/deprecated-classes': 1,
-    '@department-of-veterans-affairs/use-new-utility-classes': 1,
     '@department-of-veterans-affairs/use-workspace-imports': 1,
   },
 };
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin/lib/config/recommended.js
```diff
@@ -185,6 +185,7 @@ module.exports = {
     '@department-of-veterans-affairs/prefer-telephone-component': 1,
     '@department-of-veterans-affairs/telephone-contact-digits': 1,
     '@department-of-veterans-affairs/deprecated-classes': 1,
+    '@department-of-veterans-affairs/use-new-utility-classes': 1,
```
jamigibbs, line 188: @jhonnyoddball Do we want this set to 1 already?

packages/eslint-plugin/lib/rules/use-new-utility-classes.js
```diff
@@ -0,0 +1,95 @@
...
+  const newClass = `"${oldToNew[oldClass]}${value}"`;
+  return newClass
+}
+
+function getErrorMessage(classes) {
+  const classesArray = classes.split(" ");
+  const firstErrorClass = classesArray.find(element => element.includes('vads-u'));
+  const newClass = getNewClass(firstErrorClass);
+
+  return newClass.includes('undefined') ? MESSAGE : `${MESSAGE} Try using ${newClass} instead of "${firstErrorClass}"`;
```
jamigibbs, line 65: A bit nitpicky but instead of saying "Try using" can we just say "Use the". I think we should be clear that it must be changed and not optional in our wording.

packages/eslint-plugin/lib/config/recommended.js
jhonnyoddball: Yes, as recommended by default. But I will turn it off once it's installed in the `vets-website`

jamigibbs:
@jhonnyoddball In the description of this PR, the screenshot shows it updating a color class. Does this linting rule handle colors?

packages/eslint-plugin/lib/config/recommended.js
jamigibbs: Sounds good. Can we make sure that there is a follow-up ticket (or task added to a ticket) that reminds us to turn it on in vets-website when we're ready? Carol might be able to help figure out where that should be added.

jhonnyoddball: After thinking about it, I'm going to remove it from the recommended so we don't have to turn it off. And when we are ready we can just upgrade to dependency and manually turn it on. 
This is just in case this package is installed in another repo.

jhonnyoddball:
> 

No, it just replaces the class portion. Before the `--`

PR #978, opened 2023-10-20
Replacing color-warning-message color name

Andrew565:
## Description
Relates to [#2047](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2047).
Changes the name of color-warning-message to color-feedback-warning-background.

## Testing done
Tested locally; builds OK, and passes tests.

## Screenshots
No visual changes.

## Acceptance criteria
- [X] Name is changed

## Definition of done
- [X] Changes have been tested in IE11, if applicable
- [X] Documentation has been updated, if applicable
- [X] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)
- [X] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

### DIFF
--- a/packages/formation/sass/base/_b-variables.scss
+++ b/packages/formation/sass/base/_b-variables.scss
@@ -105,7 +105,7 @@ $color-primary-alt-darkest:  #046b99;
 $color-green-darker:         #195c27;
 $color-va-accent:            #988530; // New gold "metal" acccent
 $color-gold:                 #fdb81e;
-$color-warning-message:      #fac922; // Used for Disability benefits
+$color-feedback-warning-background:      #fac922; // Used for Disability benefits
 $color-gray-light-alt:       #eee;
 $color-va-modal-bg:          rgba($color-gray-dark, .8);
 
--- a/packages/formation/package.json
+++ b/packages/formation/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "7.0.8",
+  "version": "7.0.9",
   "description": "The VA design system",
   "keywords": [
     "va",
--- a/packages/formation/sass/_shame.scss
+++ b/packages/formation/sass/_shame.scss
@@ -110,7 +110,7 @@ body {
 }
 
 .csp-inline-patch-page-breadcrumbs {
-  background-color: $color-warning-message;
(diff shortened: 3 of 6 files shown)


packages/formation/sass/base/_b-variables.scss
```diff
@@ -105,7 +105,7 @@ $color-primary-alt-darkest:  #046b99;
 $color-green-darker:         #195c27;
 $color-va-accent:            #988530; // New gold "metal" acccent
 $color-gold:                 #fdb81e;
-$color-warning-message:      #fac922; // Used for Disability benefits
```
jamigibbs, line 108: This variable is being used in vets-website still. I'm wondering what would be the best way to do this changeover so they aren't interrupted.  

Maybe we need to keep this variable here for now and add `$color-feedback-warning-background` too? Then update vets-website after it's available?

Screenshot 2023-10-20 at 10 15 54 AM

Andrew565: For what it's worth, I'm changing this in vets-website right now, too, so there won't be any interruption. That PR should be up in the next few minutes.

jamigibbs: This repo is not directly linked to vets-website so just merging the PR won't make it available there. A new release of Formation will need to be created with this change and vets-website updated with the new version too (in your PR there would probably be easiest).

Before this PR is merged, you'll need to update the `package.json` version number and then follow [step 3 in the docs for publishing Formation](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/getting-started/common-tasks/updating-formation/).

Andrew565: Hmm, looks like you were right, I can't make the change in vets-website until after I've made that variable available in formation. So we either have to publish the new version of formation and then bring it in at the same time as the variable name change, or make both variables available in formation, bring it in and change vets-website, and then remove it from formation and publish another release. My preference would be to just publish a new formation release once, personally.

Andrew565: Version bumped, I believe the PR needs to be merged into master before I can publish.

jamigibbs (approved):

PR #984, opened 2023-11-01
Add button eslint rule for va-button and va-table

ataker:
## Description
Closes [2046 add eslint for va-button and va-table](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2046)

### `va-button`
Display eslint rule when html tag `button` or `input` with type `button` is used instead of `va-button`.

Note: tried to add a fixer as well, but found that it would break if the button text was a variable or JSX because they would already be evaluated so it wouldn't be possible to move them to the `va-button`'s `text` prop.

### `va-table`
Display eslint rule when html tag `table` is used instead of `va-button`.

Due to similar complexities, no fixer was written for this rule.

## Testing done
Local unit tests
Tested in `vets-website` repo locally

## Screenshots
Screenshot 2023-11-01 at 10 22 14
Screenshot 2023-11-01 at 11 22 53

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/eslint-plugin/lib/rules/prefer-button-component.js
+++ b/packages/eslint-plugin/lib/rules/prefer-button-component.js
@@ -3,7 +3,7 @@ const jsxAstUtils = require('jsx-ast-utils');
 const { elementType, getProp, getLiteralPropValue } = jsxAstUtils;
 
 const MESSAGE =
-  'The <va-button> Web Component should be used to instead of the button HTML element.';
+  'The <va-button> Web Component should be used instead of the button HTML element.';
 
 module.exports = {
   meta: {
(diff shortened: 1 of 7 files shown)


packages/eslint-plugin/lib/rules/prefer-button-component.js
```diff
@@ -0,0 +1,32 @@
+const jsxAstUtils = require('jsx-ast-utils');
+
+const { elementType, getProp, getLiteralPropValue } = jsxAstUtils;
+
+const MESSAGE =
+  'The <va-button> Web Component should be used to instead of the button HTML element.';
```
jamigibbs, line 6: Typo here and in the other one too: "The <va-button> Web Component should be used ~to~ instead of the button HTML element."

jamigibbs (approved):
Looks good. After this is merged, we'll just want to make sure a new release is created and vets-website updated.

CBonade (approved):

